PR #356, opened 2022-02-24
Require User Interaction To Close Suggest

evancrow:
Fixed: #341

Before we were closing the Suggest UI whenever the keyboard closed, but this PR makes it so that we check that we want the UI too close AND that the keyboard lost focused before hiding Suggest/Zero Query.


dan-alcantara-neeva:
Closing this one because it was fixed in a different way ages ago.

Issue #874, opened 2022-08-08
Add download manager UI [Was: User reports of downloads ending up app's internal storage rather than the shared Downloads folder]


darinwf:
Some users are reporting that downloaded files end up in the app's internal storage (`external_files/Android/data/com.neeva.app/files/Download/`) rather than in the shared Downloads folder.

Observed using a Pixel 6a on Android 12.

(This was reported through the Neeva Makers community)

Kokobe:
https://user-images.githubusercontent.com/20916043/191871348-ae4a9710-c30c-4894-886f-ec3b5ceb18ab.mov

Got overwriting files to work!! However, there is a limit of 100 (see `kMaxUniqueFiles` ) under 1 name, otherwise the file will just be overriden. Since most images downloaded from the Neeva images actually don't have unique names (they usually have `th_id=OIP.jpeg` as the file name).  I can see this being a potential problem for long-term users where suddenly some images they downloaded a while ago are now being overwritten. 

Now I could just increase `kMaxUniqueFiles` to 1000 because it is unlikely that a user would have downloaded 1000 images with the exact same name right? I guess it also just delays the problem. 

Any ideas on what to do? @darinwf @dan-alcantara-neeva

dan-alcantara-neeva:
Would try to figure out why we end up using the same filename for everything...

Kokobe:
Fingers crossed: https://neevaco.slack.com/archives/CGG8Z4ZSA/p1663892340007799

I'll keep it at 100 for now. Not sure how many people actually use neeva images and download images from there

Kokobe:
Closing because we no longer need a download manager UI because of the download fixes. We may need this in the future when writing to external storage is actually deprecated.

dan-alcantara-neeva:
Going to reopen this because trying to use the system downloads directory now fails after Android 13 was fully released.

PR #976, opened 2022-09-14
Add Download Callbacks and Snackbar Popups 

### LINKED ISSUE #875: Should show a toast when a download starts and finishes
I've scanned QR codes at restaurants, which often results in a PDF being downloaded with no tab.  I've missed the notification for the download multiple times, so a toast saying to look up at the notifications would be useful.

Kokobe:
Uses `weblayer` to add a callback that listens for download events (`onStarted`, `onCompleted`, `onFailed`). 

* Displays a popup notification for each respective download event. 
* For `onCompleted`, it adds an action button to the popup notification that will open the file in the Downloads app 
* Created an injectable `ActivityStarter` as a general way to safely start an activity. 
* [Clickup task: Notify user using popup when download complete](https://app.clickup.com/t/3hqfnry)
* Discussion: #874 

### How was this tested?
- [x] I tested this manually
- [x] I added automated tests

### Manual test cases
See below! (Tested and confirmed on API 29 as well)

#### `onCompleted`: Downloading an image and immediately viewing it:

https://user-images.githubusercontent.com/20916043/190792529-ea575b29-2316-4c56-8b02-b59df17a9480.mov

#### `onFailed`: Downloading an image and it failed (tested by turning on Airplane Mode)
* If the image was already downloaded, `onCompleted` will still run

https://user-images.githubusercontent.com/20916043/190792684-a2113fc9-c8d9-4060-b51d-79e1a9ed7f4f.mov

## Issues addressed
* #874 
* Fixes #875


app/src/main/java/com/neeva/app/browsing/BaseBrowserWrapper.kt
```diff
@@ -393,6 +394,7 @@ abstract class BaseBrowserWrapper internal constructor(
 
         val browser = browserFlow.value ?: throw IllegalStateException()
         registerBrowserCallbacks(browser)
+        setUpDownloads(browser)
```
dan-alcantara-neeva, line 397: I think this can fire repeatedly, meaning that you will end up setting that download callback over and over again.  Is there a reason you moved it out of the other place?

app/src/main/java/com/neeva/app/downloads/DownloadCallbackImpl.kt
```diff
@@ -0,0 +1,91 @@
...
+    private val popupModel: PopupModel,
+    private val appContext: Context
+) : DownloadCallback() {
+    override fun onInterceptDownload(
+        uri: Uri,
+        userAgent: String,
+        contentDisposition: String,
+        mimetype: String,
+        contentLength: Long
+    ): Boolean { return false }
```
dan-alcantara-neeva, line 28: Put `return false` on its own line, or just make this line be `) = false`

app/src/main/java/com/neeva/app/downloads/DownloadCallbackImpl.kt
```diff
@@ -0,0 +1,91 @@
...
+            message = appContext.resources.getString(
+                R.string.download_failed,
+                downloadFileName
+            )
+        )
+    }
+
+    override fun onDownloadCompleted(download: Download) {
+        super.onDownloadCompleted(download)
+        // Storing as separate variables so onActionPerformed lambda does not store Download object
```
dan-alcantara-neeva, line 61: Lambdas will hold onto everything that's in scope where they're created, so this will hold onto `download` unless `download` is passed along from Chromium as a `WeakReference`.  What are you seeing that encouraged you to cache the values locally?

app/src/main/java/com/neeva/app/downloads/DownloadCallbackImpl.kt
```diff
@@ -0,0 +1,91 @@
...
+                val openFileIntent = Intent()
+                    .setAction(Intent.ACTION_VIEW)
+                    .setDataAndType(Uri.fromFile(location), mimetype)
+                    .setFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+                safeStartActivityForIntent(openFileIntent)
+            }
+        )
+    }
+
+    private fun safeStartActivityForIntent(intent: Intent) {
```
dan-alcantara-neeva, line 79: Should just add a new utility function for this instead of duplicating/forking the existing copy.

app/src/main/java/com/neeva/app/downloads/DownloadCallbackImpl.kt
```diff
@@ -0,0 +1,91 @@
...
+        try {
+            appContext.startActivity(intent)
+        } catch (e: ActivityNotFoundException) {
+            popupModel.showSnackbar(appContext.getString(R.string.error_generic))
+            Log.e(TAG, "Failed to start Activity for $intent")
+        }
+    }
+
+    companion object {
+        private val TAG = DownloadCallbackImpl::class.simpleName
```
dan-alcantara-neeva, line 89: Just use `DownloadCallbackImpl` so that Proguard doesn't munge the name in logs for release builds.

app/src/main/res/values/strings.xml
```diff
@@ -809,4 +809,15 @@
 
     <!-- [NeevaScope] Sends the user to the support page  -->
     <string name="neevascope_contact_us">Contact us via Support</string>
+
+
+
+    <!-- [Downloads] Text for the snackbar popup notification that indicates the download has started for a given file. -->
+    <string name="download_started">Download started for: %1$s</string>
```
dan-alcantara-neeva, line 816: Does iOS show popup notifications?  Do they have existing strings?

dan-alcantara-neeva (changes requested):

app/src/main/java/com/neeva/app/browsing/BaseBrowserWrapper.kt
Kokobe: Originally I actually had that Download Callback setup in `registerBrowserCallbacks()`. But now I see that the implementations of `BaseBrowserWrapper` actually check if they were already registered or not. 

Moving it back to registerBrowserCallbacks!

app/src/main/java/com/neeva/app/downloads/DownloadCallbackImpl.kt
Kokobe: So when I didn't use local variables and accessed them directly in the lambda like this:

```
onActionPerformed = {
                val openFileIntent = Intent()
                    .setAction(Intent.ACTION_VIEW)
                    .setDataAndType(Uri.fromFile(download.location), download.mimeType)
                    .setFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
                safeStartActivityForIntent(openFileIntent)
            }
```
I get this error:
```
 E  FATAL EXCEPTION: main
                                                                                                    Process: com.neeva.app.debug, PID: 28169
                                                                                                    java.lang.IllegalStateException: Using Download after native destroyed
                                                                                                    	at org.chromium.weblayer_private.DownloadImpl.throwIfNativeDestroyed(DownloadImpl.java:332)
                                                                                                    	at org.chromium.weblayer_private.DownloadImpl.getLocation(DownloadImpl.java:294)
                                                                                                    	at org.chromium.weblayer_private.interfaces.IDownload$Stub.onTransact(IDownload.java:146)
                                                                                                    	at android.os.Binder.transact(Binder.java:667)
                                                                                                    	at org.chromium.weblayer_private.interfaces.IDownload$Stub$Proxy.getLocation(IDownload.java:322)
                                                                                                    	at org.chromium.weblayer.Download.getLocation(Download.java:123)
                                                                                                    	at com.neeva.app.downloads.DownloadCallbackImpl$onDownloadCompleted$1.invoke(DownloadCallbackImpl.kt:72)
                                                                                                    	at com.neeva.app.downloads.DownloadCallbackImpl$onDownloadCompleted$1.invoke(DownloadCallbackImpl.kt:66)
```

dan-alcantara-neeva: Ah, so the `Download` itself is fine, but the native `DownloadImpl` got nuked.

app/src/main/java/com/neeva/app/downloads/DownloadCallbackImpl.kt
Kokobe: Should I make a class with a static function that anybody can call like `ActivityStarter.safeStartActivityForIntent(...)` is how you would call it from anywhere?

app/src/main/java/com/neeva/app/downloads/DownloadCallbackImpl.kt
Kokobe: Ah yeah cause `Download` has a private variable `IDownload mDownloadImpl` : 

Screen Shot 2022-09-20 at 10 20 03 AM

app/src/main/res/values/strings.xml
Kokobe: iOS does not have popup notifications :( 

I guess iOS users just trust that it happened and will check their Photos or Files themselves.

app/src/main/java/com/neeva/app/downloads/DownloadCallbackImpl.kt
dan-alcantara-neeva: Debating it... Might clean up `AppNavModelImpl` if you make another injectable that allows that.  Also means `ShareSpace` wouldn't need to randomly use `AppNavModel` to do it.  Trying to think of situations where we'd want to generalize it so that you could pass in (say) an Activity context instead of the application context, but we don't have any of those currently.

Give it a shot?

Kokobe: If I made another singleton injectable or used my method above and held an `ApplicationContext`, there wouldn't be any memory leak problems because ApplicationContext only dies when the applciation dies right?

So if I used an Activity Context, rotating the screen can cause my singleton to hold onto a deleted object no? Trying it out now

Kokobe: Also, I remember we had a conversation in person about not having `DownloadCallbackImpl` be an injectable class. (right now it depends on popupModel, appContext, and my new `ActivityStarter` class). 

We said that functionally, it didn't need to be an injectable and for the tests we would have to create mock versions of it too. But now that I'm making `ActivityStarter` injectable, I feel like I should do the same for `DownloadCallbackImpl` too? WDYT?

Kokobe:
Also, I remember we had a conversation in person about not having DownloadCallbackImpl be an injectable class. (right now it depends on popupModel, appContext, and my new ActivityStarter class).

We said that functionally, it didn't need to be an injectable and for the tests we would have to create mock versions of it too. But now that I'm making ActivityStarter injectable, I feel like I should do the same for DownloadCallbackImpl too? WDYT?

app/src/main/java/com/neeva/app/downloads/DownloadCallbackImpl.kt
dan-alcantara-neeva: Roughly, yes: that's how WebLayer has been set up.  That's why we have to constantly check if the Browser and Tab objects are alive when we use them -- especially when we're doing things asynchronously and the object could have been nuked while we were processing things in the background.

That said, caching the fields is the right thing to do.  We do something similar with the `Tab` when we want to save its ID while something runs in the background.

app/src/main/java/com/neeva/app/downloads/DownloadCallbackImpl.kt
dan-alcantara-neeva: Copying some points from the offline Slack thread:

* `@ApplicationContext` is an annotation for Hilt that says we need it to inject the `Context` that is the `Application`.

* The `Application` lives as long as the process does, which means that it will outlast all of the `Activity`s that we have, which can be killed/restarted when the screen is rotated (among other ways).
  * This is why you can (e.g.) register an `ActivityLifecycleCallback` on the `Application` and be alerted when an `Activity` is created/started/stopped/etc.

* `@ActivityContext` is an annotation that says you want Hilt to inject an `Activity` as a `Context`.  You can't inject it into  a `@Singleton` because `Singleton` survives as long as the `Application` does. 

* The `Context` roughly represents the environment that runs around your code, which you use to access resources and make various system calls.  You ideally want to be using the closest context you can:
  * If you theme your `Application` differently from your `Activity`, you want to make sure that you pull the correct resources/colors/strings.
  * There are times when you don't want to be using the `Activity` `Context` because it can be killed.  If you have (say) an asynchronous process that should outlast the Activity, you want to use the `Application` `Context` instead.

dan-alcantara-neeva: Answer to this StackOverflow question is super old, but it's roughly still relevant:
https://stackoverflow.com/questions/4128589/difference-between-activity-context-and-application-context

dan-alcantara-neeva:
> Also, I remember we had a conversation in person about not having `DownloadCallbackImpl` be an injectable class. (right now it depends on `popupModel`, `appContext`, and my new `ActivityStarter` class).
> 
> We said that functionally, it didn't need to be an injectable and for the tests we would have to create mock versions of it too. But now that I'm making `ActivityStarter` injectable, I feel like I should do the same for `DownloadCallbackImpl` too? WDYT?

Still don't see a huge benefit for it, but if you're going to have the option between injecting a `DownloadCallbackImpl` and a `ActivityStarter` into the `BaseBrowserWrapper` classes, you might as well do the `DownloadCallbackImpl` to avoid piping this new field into a `DownloadCallbackImpl` you make on the fly.

Either way, you're going to have to mock _something_ for the tests.

app/src/main/java/com/neeva/app/downloads/DownloadCallbackImpl.kt
dan-alcantara-neeva: Should also toss on:
Another reason you'd normally want to use an `Activity` as the `Context` for launching another app is that you have the option of putting the new `Activity` on the same Android task stack (so they appear on top of our Activity in Android's recent apps list).  When you use `FLAG_ACTIVITY_NEW_TASK`, you tell Android to avoid doing that.  I'm mostly OK with that because it's hard to think of situations where we'd want another app to cover up their browser -- generally we're the ones sending the user to other apps.

dan-alcantara-neeva (commented):
You planning on adding tests for this?

Kokobe:
> You planning on adding tests for this?

Not sure which kind of tests I should add @dan-alcantara-neeva. Instrumentation tests wouldn't be able to connect to the internet so the only Download callback that would work is the started or failure case. 

I can write tests to make sure that the popup is being shown when each download callback is being called. Is that useful?

Kokobe:
Making a test similar to `TabCallbacksTest` for DownloadCallback

dan-alcantara-neeva:
> Making a test similar to `TabCallbacksTest` for DownloadCallback

Yeah, was expecting tests for the `DownloadCallbackImpl`, just to make sure that things were firing as expected.

app/src/test/java/com/neeva/app/browsing/DownloadCallbackTest.kt
```diff
@@ -0,0 +1,107 @@
...
+import org.mockito.Mockito.times
+import org.mockito.Mockito.verify
+import org.mockito.kotlin.any
+import org.mockito.kotlin.argumentCaptor
+import org.mockito.kotlin.doReturn
+import org.mockito.kotlin.eq
+import org.mockito.kotlin.mock
+
+@RunWith(AndroidJUnit4::class)
+class DownloadCallbackTest : BaseTest() {
```
dan-alcantara-neeva, line 30: This is a `DownloadCallbackImplTest` because you're testing the `DownloadCallbackImpl`.

app/src/test/java/com/neeva/app/browsing/DownloadCallbackTest.kt
```diff
@@ -0,0 +1,107 @@
...
+                )
+            ),
+            duration = any(),
+            onActionPerformed = onActionPerformedLambda.capture(),
+            onDismissed = any()
+        )
+
+        onActionPerformedLambda.lastValue.invoke()
+        verify(activityStarter, times(1)).safeStartActivityForIntent(
+            any()
```
dan-alcantara-neeva, line 92: Might be worth checking that the Intent is doing what you want it to?

dan-alcantara-neeva (commented):
Worth adding a check to see that the Intent has the right data, but overall looking good.

PR #977, opened 2022-09-14
Add bloom filter binary download and test

yasminyin96:
* Add downloading pipeline([flowchart](https://mermaid.live/edit#pako:eNqNU01v2kAQ_SujPQPyZww-FDXYfCRWDqGVqtg5rOyhrGJ7Le86hAL_vYttCCBo45N35r03b95qNyTmCRKX_C5psYQfXpSD-r6HMwE_nwN4pylLhq_Q7X6DbX3Ywn3o8VWecppAvCwxfhNVBguW4mtLbtA5l9AyXsK55EXbvq_bo82UCvA_Yiwk47kY7pruqCE_oVzx8g3wALgQaWFHI__BPfFThBcGjfnGewv1aqh_w5ffCAX_GuZfGzZWaaGAlMc03acksazDUhgm5LCljhvqGsUWJuFo7wzYApiEjEplVECOq0vH40PQW5h-3snJjBY3ORH_Fc7pOx5ANE9AfC4w-aLgtMbNzqM6JDX74tXMrqX1cHX3i70fTvZ5PO6D5939EmfzHut6cON-g4Z1FLtpO7hm-0aqpEMyLDPKEvXCNnt-ROQSM4yIq34TXNAqlRGJ8p2C0kry-TqPiSvLCjukKhIq0WNUvc3svOgnTPKSuAuaClUsaP7C-RGjjsTdkA_i6nbP0CzDMPuW3betu8GgQ9bE7ep9p2dZpjkYaH1Tu7N1Z9chf2oJvWfomm2ZlmNrmmM4jrX7C8YYU48)): 

1. download checksum file using `OkHttpClient`
2. download reddit filter binary file(`reddit.bin`) using `OkHttpClient`
3. use `WorkManager` to limit the download task

> Example of checksum file:
{
	"sha": "cdc5da74cb17d4fca031bb16ef1961e9775037aed9025d604d3c6d6dd3ae571b",
	"md5": "837d7499739d451b7944a230e07acddc",
        "hash": "fe7351f",
	"canon": "5aa27ef"
} 

* Add loading pipeline([flowchart](https://mermaid.live/edit#pako:eNqNVd9vmzAQ_lcs9wWkJAosv-Bh07Y00qRuqpp107bswYUjYXEwM0Yro_nfZ2MIUMhSPxBz99333XFnJ8ce8wG7eMtJvEOfl5sIybUWhAs0HL5G73fg7W8Y8cHXroZBAZ4-sSekXm_DGGgYQRHaD_0GicYqLg3pRNaibykH4mcKEUbbBmPbUTOvBYt_qMfP_4NVxoWWMlaJnAx1AtfR7xTSVt2VqRa9k-yrkEJyDqXUSoNR_iKf_Ymo0hIk2ZuFYoNHM-nnyVyAVvmSya3YAXpHGTsg6RLAUSARCB7DRLw56rhVo1IWN22nJhi3nHmQnKEz293_Qmjo51-JRhd6sQ6XZSapp3ZBSmmmEnjW_SK2J58egEpuySIo22V8JHyv1aQyLSbHbHXMKEhUZ8MAEXR_d4PCBIVRpyizM5L5h57StUb1FTXQuE_gHF9nfg09xbSct7j0mc_aaahdTxMDLg2JYJxswewZKZU0ac8PgtJXZd0z9721einnEAmaoQdQubZrPx2IsqJf7AEJhrRs92PooERksoh6aKsla6Pu1UyuIBh4jDLuXgXFasbVZ7AdFwQqsonsXhsvU2iRX1QAv4u8XAOLX6ZQXQaXkY0T0YPEA3wAfiChL6_xXMVtsGzQATbYlVsfApJSscGb6CihJBVsnUUedgVPYYDT2CcCliGRfwCHtvHaD-UgYjcgNJHGmETfZcsrjHzFbo4fsTt5NR_Z46k1dhbzxWRmO7MBzrA7tC1rZE8mzng2X4wdy1lMjwP8t6CwRtbUWtiWM59PplPLmtjHf-loHeM)): 

1. load the Bloom Filter files
2. run query

* Update `canonical_url_v3.yaml`: add 2 test urls

* Update `Checksum`: add `hash` to check the hash versioning and `canon` to check the canonicalization versioning

* Add `testRedditFilter` unit test

* Add `BloomFilterDownloaderTest`: use OkHttp's `MockWebServer` to mock out the network responses to check the downloading pipeline

* [Filter and checksum URL](https://paper.dropbox.com/doc/Cheatsheet-Data--BpcI0bIUPQDpzTCXpiaNzRMhAg-lo1Qu3SeK9yUD7vipb60i)
* [NeevaScope Android Bloom Filter download management](https://paper.dropbox.com/doc/NeevaScope-Android-Bloom-Filter-download-management--Bp8WCQ_jBOQXF~7c46A0pkv_Ag-N67DRXpthzq5U9e8Y8W3r)

### How was this tested?
- [X] I tested this manually
- [X] I added automated tests


app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,245 @@
...
+import okhttp3.OkHttpClient
+import okhttp3.Request
+
+class BloomFilterManager(
+    private val filesDir: Deferred<File>,
+    private val coroutineScope: CoroutineScope,
+    private val dispatchers: Dispatchers
+) {
+    companion object {
+        private val TAG = BloomFilterManager::class.simpleName
```
dan-alcantara-neeva, line 34: Use "BloomFilterManager".  Stops it from getting munged by Proguard when we build a release build

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,245 @@
...
+        private const val DIRECTORY_BLOOM_FILTER = "bloom_filter"
+    }
+
+    @JsonClass(generateAdapter = true)
+    data class CheckSum(
```
dan-alcantara-neeva, line 39: `Checksum`.  It's one word (not two), so there's no camelcasing.

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,245 @@
...
+        val filterUrl: URL,
+        val checksumUrl: URL,
+        var localUrl: Uri
+    )
+
+    private val reddit = BloomFilterConfiguration(
+        identifier = "reddit",
+        filterUrl = URL("https://s.neeva.co/web/neevascope/v1/reddit.bin"),
+        checksumUrl = URL("https://s.neeva.co/web/neevascope/v1/reddit_latest.json"),
+        // file:///data/user/0/com.neeva.app.debug/cache/bloom_filter/reddit.bin
```
dan-alcantara-neeva, line 55: What's the comment for?

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,245 @@
...
+        private const val DIRECTORY_BLOOM_FILTER = "bloom_filter"
+    }
+
+    @JsonClass(generateAdapter = true)
+    data class CheckSum(
+        @Json(name = "sha") val sha: String,
+        @Json(name = "md5") val md5: String
+    )
+
+    data class BloomFilterConfiguration(
```
dan-alcantara-neeva, line 44: Add a `kdoc` to explain why this data class exists.  Looks like you're going to have one configuration for each type of bloom filter, with the only one currently being the one for Reddit?

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,245 @@
...
+
+    private val reddit = BloomFilterConfiguration(
+        identifier = "reddit",
+        filterUrl = URL("https://s.neeva.co/web/neevascope/v1/reddit.bin"),
+        checksumUrl = URL("https://s.neeva.co/web/neevascope/v1/reddit_latest.json"),
+        // file:///data/user/0/com.neeva.app.debug/cache/bloom_filter/reddit.bin
+        localUrl = Uri.EMPTY
+    )
+
+    val bloomfilter = BloomFilter()
```
dan-alcantara-neeva, line 59: `val bloomFilter`
Bloom Filter is two words

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        return bloomfilter.mayContain(key)
+    }
+
+    private suspend fun getBloomFilterDirectory(): File {
+        return File(filesDir.await(), DIRECTORY_BLOOM_FILTER)
+    }
+
+    suspend fun getFilterBinaryFile(name: String): File {
```
dan-alcantara-neeva, line 74: Could you just inline the `getBloomFilterDirectory` function?

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,245 @@
...
+        checksumUrl = URL("https://s.neeva.co/web/neevascope/v1/reddit_latest.json"),
+        // file:///data/user/0/com.neeva.app.debug/cache/bloom_filter/reddit.bin
+        localUrl = Uri.EMPTY
+    )
+
+    val bloomfilter = BloomFilter()
+
+    fun contains(key: String): Boolean? {
+        if (bloomfilter.filter == null) {
+            load()
```
dan-alcantara-neeva, line 63: So this is always going to fail until it finishes loading?

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    suspend fun getFilterBinaryFile(name: String): File {
+        return File(getBloomFilterDirectory(), name)
+    }
+
+    // Download and save the binary file locally
+    fun load() {
+        // check url validity
+        if (
```
dan-alcantara-neeva, line 81: `filterUrl` and `checksumUrl` are constants, right?  Should you use `assert` instead?  That way if a developer accidentally changes the URLs to something wrong it'll crash instead of logging a failure.

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,245 @@
...
+            return
+        }
+
+        coroutineScope.launch(dispatchers.io) {
+            val file = getFilterBinaryFile("reddit.bin")
+
+            // Load checksum file
+            val checksumRequest = Request.Builder().url(reddit.checksumUrl).build()
+            try {
+                OkHttpClient().newCall(checksumRequest).execute().body?.let {
```
dan-alcantara-neeva, line 95: 1. Don't keep creating `OkHttpClient`s: make one and cache it.  Is it OK to use an `OkHttpClient` that has the user's login tokens (if they're logged in)?  If so, then I can make a global one because I'm currently mucking with a bunch of these for the preview tokens.
2. `execute()` returns something that needs to be closed when you're done with it.  You can use the `use` function to make sure it gets closed correctly: `newCall(checksumRequest).execute().use {`

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,245 @@
...
+
+                    // Check if the existing local filter file matches new checksum
+                    if (reddit.localUrl != Uri.EMPTY && verifyFile(reddit.localUrl, checksum)) {
+                        BloomFilter().loadFilter(reddit.localUrl)
+                        return@launch
+                    }
+
+                    // Download new binary file
+                    val request = Request.Builder().url(reddit.filterUrl).build()
+                    try {
```
dan-alcantara-neeva, line 111: You're already in a try block; you don't need to put another try.

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        coroutineScope.launch(dispatchers.io) {
+            val file = getFilterBinaryFile("reddit.bin")
+
+            // Load checksum file
+            val checksumRequest = Request.Builder().url(reddit.checksumUrl).build()
+            try {
+                OkHttpClient().newCall(checksumRequest).execute().body?.let {
+                    val checksum = try {
```
dan-alcantara-neeva, line 96: You're already in a try block; you don't need to nest them.  If you want to know which call in your function failed, split them apart.

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,245 @@
...
+        ) {
+            Log.e(TAG, "Invalid filter or checksum file")
+            return
+        }
+
+        coroutineScope.launch(dispatchers.io) {
+            val file = getFilterBinaryFile("reddit.bin")
+
+            // Load checksum file
+            val checksumRequest = Request.Builder().url(reddit.checksumUrl).build()
```
dan-alcantara-neeva, line 93: The way you've set this up, the app will keep trying to download the large Bloom Filter file over and over again until you successfully parse them.  If the backend is broken and changes its file format, or if we get output that we don't expect, or if the download gets interrupted, etc... you will be forcing the user to try downloading a large file over and over again.

We should be making an attempt to download these files at most once a day or so.  We should also make sure that the user is on an unmetered network connection to avoid destroying their mobile data allowances.  Can you sync with @EdwardJTL and @darinwf to see what behavior would be acceptable?

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,245 @@
...
+            getCheckSumFromFile(MessageDigest.getInstance("SHA-256"), fileUrl),
+            getCheckSumFromFile(MessageDigest.getInstance("MD5"), fileUrl)
+        )
+
+        return checksum.sha == fileHash.sha && checksum.md5 == fileHash.md5
+    }
+
+    // Calculate checksum hash for a file
+    // ref: https://www.geeksforgeeks.org/how-to-generate-md5-checksum-for-files-in-java/
+    private fun getCheckSumFromFile(digest: MessageDigest, fileUrl: Uri): String {
```
dan-alcantara-neeva, line 148: `getChecksumFromFile`

app/src/test/java/com/neeva/app/neevascope/BloomFilterTest.kt
```diff
@@ -101,12 +102,37 @@ class BloomFilterTest : BaseTest() {
...
+    fun testRedditFilter() {
+        val filter = BloomFilter()
+        val binFile = File("src/test/resources/neevascope/reddit.bin")
+        filter.loadFilter(Uri.fromFile(binFile))
+
+        val testFile = File("src/test/resources/neevascope/reddit_test.txt")
+        val testUrls = testFile.readLines().filterNot { it.isBlank() }
+
+        val knownErrors = listOf(
+            // Backend upper case these percent encodings
```
dan-alcantara-neeva, line 124: Are these known errors in our implementation or in the BIN file produced by the backend?

dan-alcantara-neeva (changes requested):

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
yasminyin96: Oh, I forgot to remove this comment. I will remove this line.

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
yasminyin96: The file should be only downloaded through wifi and avoids cellular. I saw there is `ConnectivityManager` used in `NeevaUser` and `.isNetworkTypeWifi` which can check the network type, but it seems to be deprecated. I'm not sure how to disable the cellular.

The file should be only downloaded when users first browse the website and then the file will be saved locally. 
But when I quit the app and restart it, it will download the file again.

dan-alcantara-neeva: Any time you're dealing with network connectivity, you need to keep the user's data plan in mind -- especially when you're downloading files (how big is the bloom filter file, anyway?  is it compressed?  should it be?).  The last thing you want to do is burn through their entire data plan because of a coding error or oversight.  Forcing the user to potentially download files over and over again when the app starts up is one of these situations.

1. You shouldn't disable cellular; you should avoid downloading anything if they're on cellular.  Sync with @Kokobe about how he did it.

2. There are many reasons your pipeline can fail and force you to redownload the files unnecessarily.  Off the top of my head:
  * If the backend mistakenly serves a checksum file that doesn't match the bloom filter file, you will keep downloading the bloom filter file because the checksums won't match.
  * If some network error prevents you from completely downloading the file (say the network gets cut off), you will keep trying until you succeed because you only check to see if there is a bloom filter file set up on line 62.
  * If there is an error processing the bloom filter file while you try to save it on line 114 (say because the user has no more storage on their device), you will try to download the file again.

To move forward, you should:
1. Look into what the non-deprecated methods for checking for **unmetered** connections are (ask @Kokobe).  Even wi-fi connections can be metered and can result in the user getting charged.
2. Sync with @EdwardJTL about how iOS handles their bloom filter downloading pipeline.  If they're not limiting the number of times these two files are being downloaded, they should be.
3. Make a flowchart that shows how everything can go and what should be done in those situations.  See either [this flowchart](https://paper.dropbox.com/doc/createOrSwitchToTab-behavior--BlQpTLcJ7qbCfDW0k96cRqjmAg-qvXWgOhIhtrLKkr62NFOS) or [this flowchart](https://mermaid.live/edit#pako:eNp9k91O4zAQhV9lZC4ACbiucgGCpqUtCGn5UaVNuDDOmFo4nsh2qo2avjtO0yZlKUSylBx9czJzbK-YoAxZxN4tLxbwHKcGwnOdpOwBcclfHNoLiV4sTk5T9grn55dwczJ14BdoETh01DN9oLk6betvGrCu0NUwXE24A-f5m0aYq7G6Wu8zhmqIkweCxmNqJIEnEBq5fW2xYW81Sq4L0prmodcCLYQlyeauq_1Toq2-1DX24yQmc-xb147dYqPNSLebJkf_BBZekXG7Hm83JnMlFQzJGBReLZWverKGyW_ubX1Hh9S4B-XAkIfvrhY195jVME2GP7uFrPpGa5it2lDgEV0RFIRFM4q1ZLsxZn2Gd4etZ11a94lD3-9G6ZR5hyXXKoP_frQtdb4KOxuDVFpHR0KIfXl8WJ4clqdbWQ4G-_LdYfl-Kw_kRmZnLEebc5WF47xqlJSFU5pjyqLwmqHkpfYpS806oGWRhaxHmfJkWSS5dnjGeOnpqTKCRd6WuINixcPtyDuq4OYv0e57_QmxORPr).
4. Put this into a doc so that we can make sure that we've covered all of our bases.

yasminyin96: Got it, thank you so much.

EdwardJTL: on iOS, we download the binary file with a url session with these restrictions
```
// Disable downloading over cellular and restricted networks
        let config = URLSessionConfiguration.default
        config.allowsCellularAccess = false
        config.allowsExpensiveNetworkAccess = false
        config.allowsConstrainedNetworkAccess = false
        config.waitsForConnectivity = true
        return URLSession(configuration: config, delegate: nil, delegateQueue: downloadQueue)
```

EdwardJTL: we save the bin file to disk, and only download if there's no valid bin file in the file system.
a description of the code flow is here: https://paper.dropbox.com/doc/NeevaScope-Backlinks-Launch-Doc--Bpjj9doRh~EB20xQ9ikKnFBZAg-k00a4k8TN5luw1M58gduj
no flow chart sorry :(

Kokobe: Reposting from our slack convo: 
A non-deprecated way to check for unmetered connections is: https://developer.android.com/training/monitoring-device-state/connectivity-status-type

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,245 @@
...
+) {
+    companion object {
+        private val TAG = BloomFilterManager::class.simpleName
+        private const val DIRECTORY_BLOOM_FILTER = "bloom_filter"
+    }
+
+    @JsonClass(generateAdapter = true)
+    data class CheckSum(
+        @Json(name = "sha") val sha: String,
+        @Json(name = "md5") val md5: String
```
EdwardJTL, line 41: we are also now checking backend tool versions. see
https://github.com/neevaco/neeva-ios/pull/4372

you can just hard code these values on the client side

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,200 @@
...
+
+    private fun getFilterBinaryFile(name: String): File {
+        val file = File(appContext.cacheDir, DIRECTORY_BLOOM_FILTER)
+        file.mkdir()
+        return File(file, name)
+    }
+}
+
+// ref: https://github.com/neevaco/neeva/blob/8907317c2ce8851599baa1ca4267724aa8687665/base/collections/bloom/bloom.go
+class BloomFilter {
```
dan-alcantara-neeva, line 124: Move this out to its own class

app/src/main/java/com/neeva/app/neevascope/BloomFilterDownloadWorker.kt
```diff
@@ -0,0 +1,144 @@
...
+            // Download and load checksum file
+            val checksumRequest = Request.Builder().url(reddit.checksumUrl).build()
+            val checksum = okHttpClient.newCall(checksumRequest).execute().use { response ->
+                response.body?.string()?.let { body ->
+                    loadChecksum(body)
+                }
+            } ?: return@withContext
+
+            // Check if the existing local filter file matches new checksum
+            // If match, don't update the existing local filter fil
```
dan-alcantara-neeva, line 80: local filter **file**

app/src/main/java/com/neeva/app/neevascope/BloomFilterDownloadWorker.kt
```diff
@@ -0,0 +1,144 @@
...
+
+    @JsonClass(generateAdapter = true)
+    data class Checksum(
+        @Json(name = "sha") val sha: String,
+        @Json(name = "md5") val md5: String,
+        @Json(name = "hash") val hash: String,
+        @Json(name = "canon") val canon: String
+    ) {
+        companion object {
+            const val hasher = "fe7351f"
```
dan-alcantara-neeva, line 45: `const val` names should be all caps.  Should also specify what these actually mean with comments.

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,200 @@
...
+            bloomFilter.loadFilter(reddit.localUri)
+            loading = false
+        } catch (e: Exception) {
+            Log.e(TAG, "Failed to load filter ", e)
+        }
+    }
+
+    private fun getFilterBinaryFile(name: String): File {
+        val file = File(appContext.cacheDir, DIRECTORY_BLOOM_FILTER)
+        file.mkdir()
```
dan-alcantara-neeva, line 118: 1. Should you check if the directory exists before you try to create it?
2. You're not checking for any exceptions that get fired.  You need to add `try` and `catch` blocks and decide how to gracefully fail.  Probably put that in the design doc.

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,200 @@
...
+                load()
+            }
+            return null
+        }
+
+        return bloomFilter.mayContain(key)
+    }
+
+    // Loading pipeline: loading the Bloom Filter files and running queries against the loaded Bloom Filters
+    private fun load() {
```
dan-alcantara-neeva, line 92: Make this a suspend function so that it is explicitly telling the caller that it needs to be used inside of a `coroutineScope`

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,200 @@
...
+        if (loading) { return }
+
+        when (workManager.getWorkInfoByIdLiveData(downloadWorkerRequest.id).value?.state) {
+            WorkInfo.State.ENQUEUED -> {
+                return
+            }
+            else -> {
```
dan-alcantara-neeva, line 99: 1. Explicitly list out the other `WorkInfo.State` possibilities in a when loop instead of relying on `else`.  Otherwise it's hard to know why you're just funneling everything into re-enqueueing it.

2. If you're doing a periodic work request, why would you need to re-enqueue it here on line 100?

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,200 @@
...
+                workManager.enqueue(downloadWorkerRequest)
+            }
+        }
+
+        if (!reddit.localUri.toFile().exists()) { return }
```
dan-alcantara-neeva, line 104: If you're putting the return on the same line, you can skip the braces.

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,200 @@
...
+
+        loading = true
```
dan-alcantara-neeva, line 106: 1. This needs to be put right next to where you check for whether `loading == true`.  Otherwise, any early exits you included on line 97 and 104 will cause the variable to never be set.

2. This shouldn't be put inside of the `coroutineScope`.  The `IO` `Dispatcher` can run multiple coroutines in parallel, meaning that you can get into a race condition where you're checking a stale value of `load`.  Instead make sure it's checked and set before you even try to `launch` the coroutine (so if `loading == false`, launch the coroutine on line 82).

3. This shouldn't be a `loading` field: it should be a `loadAttempted` field so that the app doesn't try to load the file again until the next time it starts.  Otherwise, if you have a screwed up Bloom Filter file, you'll end up trying to load it every time you want to check the Bloom Filter (so on every page load).

dan-alcantara-neeva (changes requested):
Taking a first pass.  We really should finish the doc before you write code out, though.

app/src/main/java/com/neeva/app/neevascope/BloomFilterDownloadWorker.kt
```diff
@@ -0,0 +1,148 @@
...
+                    }
+                }
+            }
+
+            // If the filter file matches checksum, rename temp file
+            if (verifyFile(Uri.fromFile(tempFile), checksum)) {
+                if (bloomFilterFile.exists()) { bloomFilterFile.delete() }
+                tempFile.renameTo(bloomFilterFile)
+            } else {
+                return@withContext
```
dan-alcantara-neeva, line 105: No need for this else clause; there's no further code after this so it'll return anyway.

app/src/main/java/com/neeva/app/neevascope/BloomFilterDownloadWorker.kt
```diff
@@ -0,0 +1,148 @@
...
+    // ref: https://www.geeksforgeeks.org/how-to-generate-md5-checksum-for-files-in-java/
+    private fun getChecksumFromFile(digest: MessageDigest, fileUrl: Uri): String {
+        val file = fileUrl.toFile().inputStream()
+        val byteArray = updateDigest(digest, file).digest()
+        file.close()
+
+        return Hex.bytesToStringLowercase(byteArray)
+    }
+
+    private fun updateDigest(digest: MessageDigest, data: InputStream): MessageDigest {
```
dan-alcantara-neeva, line 138: Can you inline this function into `getChecksumFromFile`?

app/src/main/java/com/neeva/app/neevascope/BloomFilterDownloadWorker.kt
```diff
@@ -0,0 +1,148 @@
...
+        val size = 1024
+        val byteArray = ByteArray(size)
+        var read = data.read(byteArray, 0, size)
```
dan-alcantara-neeva, line 141: What's `var read` here?  The number of bytes read?  Can you rename this to `numBytesRead` so that it doesn't get conflated with `data.read`?

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,156 @@
...
+            sharedPreferencesModel = sharedPreferencesModel,
+            value = workerRequest.id.toString()
+        )
+
+        return workerRequest
+    }
+
+    private suspend fun getFilterBinaryFile(name: String): File {
+        val file = File(appContext.cacheDir, DIRECTORY_BLOOM_FILTER)
+        if (!file.exists()) { file.mkdir() }
```
dan-alcantara-neeva, line 153: Don't need the `{` and `}`

dan-alcantara-neeva (changes requested):

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -0,0 +1,156 @@
...
+                createWorkerRequest()
+            )
+        } else {
+            val workInfo =
+                workManager.getWorkInfoById(UUID.fromString(workerRequestId)).get()
+
+            /**
+             * isFinished() returns true if this State is considered finished.
+             * @return True for SUCCEEDED, FAILED, and CANCELLED states
+             * Note that PeriodicWorkRequests will never enter this state (they will simply go back to ENQUEUED and be eligible to run again)
```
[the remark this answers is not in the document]
yasminyin96: 1. `this` State means the State of the `WorkRequest`. Yes, it's the comment for `isFinished` in `WorkInfo.java`
2. Yes, it passed. I will break the line at 100 chars

app/src/main/java/com/neeva/app/neevascope/BloomFilter.kt
```diff
@@ -45,19 +50,20 @@ class BloomFilter {
...
-
-        var h1 = (keyHash shr 31).toUInt()
-        val h2 = (keyHash and 0xFFFF_FFFFu).toUInt()
-        val numBits = (filter.dataByteArray.size * 8).toUInt()
-        for (i in (0).toUInt() until filter.kUInt) {
-            val bit = (h1 % numBits).toInt()
-            h1 += h2
-            if (filter.dataByteArray[(bit / 8)] and (1 shl (bit % 8)).toByte() == (0).toByte()) {
-                return false
+        filter?.let {
```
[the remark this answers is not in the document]
yasminyin96: I tried to make the set private `private var _filter: Filter? = null` `val filter: Filter? get() = _filter`
but it always shows an error "Smart cast to 'BloomFilter.Filter' is impossible, because 'filter' is a property that has open or custom getter" and needs to add non-null assesrted `!!` when I use `filter` here.

app/src/main/java/com/neeva/app/neevascope/BloomFilter.kt
```diff
@@ -29,7 +29,8 @@ class BloomFilter {
         val kUInt: UInt = k.toUInt()
     }
 
-    var filter: Filter? = null
+    private var _filter: Filter? = null
+    val filter: Filter? get() = _filter
```
dan-alcantara-neeva, line 33: You should be able to just do something like
```
var filter: Filter? private set
```

app/src/main/java/com/neeva/app/neevascope/BloomFilterDownloadWorker.kt
```diff
@@ -128,21 +147,22 @@ class BloomFilterDownloadWorker(
...
 
-    private fun updateDigest(digest: MessageDigest, data: InputStream): MessageDigest {
-        val size = 1024
-        val byteArray = ByteArray(size)
-        var read = data.read(byteArray, 0, size)
-        while (read > -1) {
-            digest.update(byteArray, 0, read)
-            read = data.read(byteArray, 0, size)
+            val digestBytes = digest.digest()
+            return Hex.bytesToStringLowercase(digestBytes)
```
dan-alcantara-neeva, line 162: Can you specify which Exceptions you're trying to catch here?  It's good to narrow it down to show that you are expecting specific things to go wrong.  If a crash is logged that we're not expecting, then we know something else is going wrong and we need to revisit our assumptions.

app/src/main/java/com/neeva/app/neevascope/BloomFilterDownloadWorker.kt
```diff
@@ -71,50 +79,61 @@ class BloomFilterDownloadWorker(
...
+            okHttpClient.newCall(request).execute()
+        } catch (e: Exception) {
             Log.e(TAG, "Failed to download checksum or filter file ", e)
+            null
         }
     }
 
     private fun loadChecksum(checksumFile: String): Checksum? {
         val moshi = Moshi.Builder().build()
         val jsonAdapter = moshi.adapter(Checksum::class.java)
```
dan-alcantara-neeva, line 130: Can't this `adapter` function also throw an exception?  Should live inside the `try` with the other things.

dan-alcantara-neeva (changes requested):

app/src/main/java/com/neeva/app/neevascope/BloomFilter.kt
```diff
@@ -36,27 +42,30 @@ class BloomFilter {
...
     fun mayContain(key: String): Boolean {
-        val keyHash = fnv1aHash(key.toByteArray())
+        if (filter == null) return false
 
+        val keyHash = fnv1aHash(key.toByteArray())
         var h1 = (keyHash shr 31).toUInt()
         val h2 = (keyHash and 0xFFFF_FFFFu).toUInt()
-        val numBits = (filter.dataByteArray.size * 8).toUInt()
-        for (i in (0).toUInt() until filter.kUInt) {
+        val numBits = (filter!!.dataByteArray.size * 8).toUInt()
```
dan-alcantara-neeva, line 60: The !! you're using now is because you can change `filter`.  If you're going to get stuck using `!!`, you should just go back to using the `let` like you had it before.

dan-alcantara-neeva (changes requested):

dan-alcantara-neeva (commented):
PR seems to do what it needs to...  @Kokobe reminded me that you need to write tests for the pipeline, though.  Would help to make sure that you're testing each of the codepaths correctly.  You'll want to look into OkHttp's `MockWebServer` to mock out the network responses:

`https://github.com/square/okhttp/tree/master/mockwebserver`

app/src/test/java/com/neeva/app/neevascope/BloomFilterDownloaderTest.kt
```diff
@@ -0,0 +1,162 @@
...
+
+        expectThat(fileSha)
+            .isEqualTo("9b5451084d26e9869a4f372bb3d2ab149d090424b30205658be292794b7ed608")
+        expectThat(fileMd5).isEqualTo("70d2143e042de8b27000450144bf2ea1")
+    }
+
+    private fun initializeServer(checksumResponse: String, bloomFilterResponse: String) {
+        server = MockWebServer()
+
+        val dispatcher = object : Dispatcher() {
```
dan-alcantara-neeva, line 86: Huh, neat.  Didn't know you could do it this way

app/src/test/java/com/neeva/app/neevascope/BloomFilterDownloaderTest.kt
```diff
@@ -0,0 +1,162 @@
...
+
+        expectThat(file.exists()).isTrue()
+        // The bloom filter is loaded from downloaded filter file.
+        bloomFilter.loadFilter(Uri.fromFile(file))
+
+        // The bloom filter should not be null.
+        expectThat(bloomFilter.filter).isNotNull()
+    }
+
+    @Test
```
dan-alcantara-neeva, line 136: Can you add a test that has a checksum file that doesn't match the bloom filter file?

dan-alcantara-neeva (approved):
Seems good to go...  Could you wait until tomorrow morning until landing this?  Want to cut a new release tonight and want to avoid putting this in in case we missed something.

yasminyin96:
> Seems good to go... Could you wait until tomorrow morning until landing this? Want to cut a new release tonight and want to avoid putting this in in case we missed something.

Sure.

PR #983, opened 2022-09-19
Add page load event

chung-lai:
* Add `PageLoad` event (no sensitive information attached) to understand frequency of page loading of user

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

### Test case

1. it logs every time a navigation happens and make sure the right attribute is added to the log


chung-lai:
> Don't think existing tests cover any of the changes you added

Ah, any existing unit test/automation that will go through a navigation flow that I can hook it up and check if this clientLogger is logging the right thing?

app/src/main/java/com/neeva/app/browsing/TabCallbacks.kt
```diff
@@ -123,6 +126,7 @@ class TabCallbacks(
             val isRestoringState = browserFlow.value?.isRestoringPreviousState == true
             if (!isRestoringState) {
                 visitToCommit = Visit(timestamp = Date())
+                clientLogger?.logCounter(LogConfig.Interaction.BROWSER_PAGE_LOAD, null)
```
dan-alcantara-neeva, line 129: 1. This can fire multiple times, if the page is triggering a bunch of navigations.  You want to try and filter any of them out?
2. Do you want to record all navigations?  This log you've added will also trigger if the user is (say) downloading file or the browser sends the user to another app.  You might want to use `commitVisit` instead.

app/src/main/java/com/neeva/app/logging/LogConfig.kt
```diff
@@ -53,14 +55,18 @@ class LogConfig {
...
+        BROWSER_PAGE_LOAD("PageLoad", Category.BROWSING)
     }
 
     enum class Category(val categoryName: String) {
         UI("UI"),
         SUGGESTIONS("Suggestion"),
         STABILITY("Stability"),
-        FIRST_RUN("FirstRun")
+        FIRST_RUN("FirstRun"),
+        BROWSING("Browsing"),
```
dan-alcantara-neeva, line 69: Can these be alphabetized?

app/src/main/java/com/neeva/app/browsing/TabCallbacks.kt
chung-lai: Oh, yes! We should only record it when user initiates a navigation so `commitVisit` seems to be better

dan-alcantara-neeva: Hrm, this might be more difficult to track if you want just user-initiated navigations.  I think this pipeline triggers for all navigations -- including those triggered by the page itself.  Worth checking to see what logs are made and what flags are available on the `navigation` that could give you an idea if it's actually user-driven.

dan-alcantara-neeva:
> > Don't think existing tests cover any of the changes you added
> 
> Ah, any existing unit test/automation that will go through a navigation flow that I can hook it up and check if this clientLogger is logging the right thing?

Worth creating a new one to do the test.  Maybe an instrumentation test, given that this thing crosses so many classes?  Something in the `logging` package?  Would need a test that does a page load and make sure that the logger tried to send something out like how `ClientLoggerSettingTest` checks, but I'm not sure how to make sure the right one gets fired.

chung-lai:
> > > Don't think existing tests cover any of the changes you added
> > 
> > 
> > Ah, any existing unit test/automation that will go through a navigation flow that I can hook it up and check if this clientLogger is logging the right thing?
> 
> Worth creating a new one to do the test. Maybe an instrumentation test, given that this thing crosses so many classes? Something in the `logging` package? Would need a test that does a page load and make sure that the logger tried to send something out like how `ClientLoggerSettingTest` checks, but I'm not sure how to make sure the right one gets fired.

Added test for logging. I'm thinking whether we should have a dedicated test suite for logging which kind of mimic other instrumentation tests. (e.g., we want to check the suggestion logging by reusing the suggestion test) Or we can add the logging check in existing instrument test to verify if those events got fired. Any feedback is welcome.

dan-alcantara-neeva:
> > > > Don't think existing tests cover any of the changes you added
> > > 
> > > 
> > > Ah, any existing unit test/automation that will go through a navigation flow that I can hook it up and check if this clientLogger is logging the right thing?
> > 
> > 
> > Worth creating a new one to do the test. Maybe an instrumentation test, given that this thing crosses so many classes? Something in the `logging` package? Would need a test that does a page load and make sure that the logger tried to send something out like how `ClientLoggerSettingTest` checks, but I'm not sure how to make sure the right one gets fired.
> 
> Added test for logging. I'm thinking whether we should have a dedicated test suite for logging which kind of mimic other instrumentation tests. (e.g., we want to check the suggestion logging by reusing the suggestion test) Or we can add the logging check in existing instrument test to verify if those events got fired. Any feedback is welcome.

Putting a logging check in an existing instrumentation test seems fine, if you can pick the right ones.

chung-lai:
@dan-alcantara-neeva, Sounds good, will start with this and add correspondingly in a separate PR.

dan-alcantara-neeva (approved):

PR #991, opened 2022-09-22
Fetch play install referrer

chung-lai:
* Add the install referrer library to gradle
* initiate connection to play api for requesting install referrer for tracking
* only fetch this information once per user

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests

**Test case**

* Load on a device and make sure the referrer url is filled


app/src/main/java/com/neeva/app/NeevaActivity.kt
```diff
@@ -301,6 +310,89 @@ class NeevaActivity : AppCompatActivity(), ActivityCallbacks {
         clientLogger.logCounter(LogConfig.Interaction.APP_ENTER_FOREGROUND, null)
         updateWidgets()
         webLayerModel.currentBrowser.reregisterActiveTabIfNecessary()
+
+        lifecycleScope.launchWhenResumed {
```
dan-alcantara-neeva, line 314: Given that this is a `launchWhenResumed` call, it will only fire once when the `lifecycleScope` has hit the `resumed` state.  However, because you're putting this call in `onStart`, you're effectively making the app reschedule this check every single time the app is brought to the foreground.  You should stick it with the other calls in `onCreate`.

app/src/main/java/com/neeva/app/NeevaActivity.kt
```diff
@@ -301,6 +310,89 @@ class NeevaActivity : AppCompatActivity(), ActivityCallbacks {
...
+            if (!SharedPrefFolder.App.RequestedInstallReferrer.get(sharedPreferencesModel)) {
+                requestInstallReferrer()
+                SharedPrefFolder.App.RequestedInstallReferrer.set(
+                    sharedPreferencesModel, true
+                )
+            }
+        }
+    }
+
+    private fun requestInstallReferrer() {
```
dan-alcantara-neeva, line 324: This is a lot of code to be sticking in the Activity.  Can you put it in the `NeevaActivityViewModel`?

app/src/main/java/com/neeva/app/NeevaActivity.kt
```diff
@@ -301,6 +310,89 @@ class NeevaActivity : AppCompatActivity(), ActivityCallbacks {
...
+                                        Optional.presentIfNotNull(
+                                            LogConfig.FirstRunAttributes
+                                                .INSTALL_REFERRER.attributeName
+                                        ),
+                                        Optional.presentIfNotNull(referrerUrl)
+                                    )
+                                )
+                            )
+                        } catch (e: RemoteException) {
+                            e.printStackTrace()
```
dan-alcantara-neeva, line 355: Use the Log class instead of a naked `printStackTrace` call.

`Log.e(TAG, "message", e)`

app/src/main/java/com/neeva/app/NeevaActivity.kt
```diff
@@ -301,6 +310,89 @@ class NeevaActivity : AppCompatActivity(), ActivityCallbacks {
         clientLogger.logCounter(LogConfig.Interaction.APP_ENTER_FOREGROUND, null)
         updateWidgets()
         webLayerModel.currentBrowser.reregisterActiveTabIfNecessary()
+
+        lifecycleScope.launchWhenResumed {
+            if (!SharedPrefFolder.App.RequestedInstallReferrer.get(sharedPreferencesModel)) {
+                requestInstallReferrer()
+                SharedPrefFolder.App.RequestedInstallReferrer.set(
```
dan-alcantara-neeva, line 317: Should this value be set if you fail to perform the request?  Because the request is fired asynchronously, you're setting the SharedPreference before it has a chance to complete.

app/src/main/java/com/neeva/app/NeevaActivity.kt
```diff
@@ -301,6 +310,89 @@ class NeevaActivity : AppCompatActivity(), ActivityCallbacks {
...
+                                    Optional.presentIfNotNull("SERVICE_UNAVAILABLE")
+                                )
+                            )
+                        )
+                    }
+                }
+            }
+
+            override fun onInstallReferrerServiceDisconnected() {
+                // Try to restart the connection on the next request to
```
dan-alcantara-neeva, line 392: Is this a comment copied from somewhere?  Don't know if you ever try to make another request because you gate it with a `SharedPreference` check.

dan-alcantara-neeva (changes requested):

app/src/main/java/com/neeva/app/NeevaActivity.kt
chung-lai: From looking at the response, I would be ok to do this once per user to start with. If we notice it fails frequently (which we have logging) and will look into whether we want to retry

dan-alcantara-neeva (approved):

dan-alcantara-neeva:
Squashing and merging based on offline chat with @chung-lai

Issue #1002, opened 2022-09-26
Crash: com.neeva.app.cardgrid.archived.ProcessedArchivedTabs.getTabData


dan-alcantara-neeva:
Crash on 0.84.0 likely caused by the asynchronous fetches in Paging and a lack of boundary condition check when getting the data out.

Introduced by me in https://github.com/neevaco/neeva-android/pull/959/files

```
Exception java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
  at androidx.paging.PagePresenter.checkIndex (PagePresenter.java:9)
  at androidx.paging.PagingDataDiffer.get (PagingDataDiffer.java:9)
  at androidx.paging.compose.LazyPagingItems.get (LazyPagingItems.java:9)
  at com.neeva.app.cardgrid.archived.ProcessedArchivedTabs.getTabData (ProcessedArchivedTabs.java:10)
  at com.neeva.app.cardgrid.archived.ArchivedTabGridKt$ArchivedTabGrid$5$5$3.invoke (ArchivedTabGridKt.java:10)
  at com.neeva.app.cardgrid.archived.ArchivedTabGridKt$ArchivedTabGrid$5$5$3.invoke (ArchivedTabGridKt.java:10)
  at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke (ComposableLambdaImpl.java)
  at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke (ComposableLambdaImpl.java)
  at androidx.compose.foundation.lazy.grid.LazyGridItemsSnapshot.Item (LazyGridItemsSnapshot.java:5)
  at androidx.compose.foundation.lazy.grid.LazyGridItemProviderImpl.Item (LazyGridItemProviderImpl.java)
  at androidx.compose.foundation.lazy.layout.LazyLayoutItemContentFactory$CachedItemContent$createContentLambda$1$1.invoke (LazyLayoutItemContentFactory.java:1)
  at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke (ComposableLambdaImpl.java)
  at androidx.compose.runtime.CompositionLocalKt.CompositionLocalProvider (CompositionLocalKt.java)
  at androidx.compose.runtime.saveable.SaveableStateHolderImpl.SaveableStateProvider (SaveableStateHolderImpl.java:3)
[...]

Issue #1003, opened 2022-09-26
Crash: com.neeva.app.cookiecutter.ScriptInjectionManager.initializeMessagePassing


dan-alcantara-neeva:
Showing up in 0.84.0 and 0.85.0, which isn't great because 0.85.0 is still beta-only as of this morning.

```
Exception java.lang.IllegalArgumentException:
  at org.chromium.weblayer_private.TabImpl.registerWebMessageCallback (TabImpl.java:1022)
  at org.chromium.weblayer_private.interfaces.ITab$Stub.onTransact (ITab.java:345)
  at android.os.Binder.transact (Binder.java:1164)
  at org.chromium.weblayer_private.interfaces.ITab$Stub$Proxy.registerWebMessageCallback (ITab.java)
  at org.chromium.weblayer.Tab.registerWebMessageCallback (Tab.java)
  at com.neeva.app.cookiecutter.ScriptInjectionManager.initializeMessagePassing (ScriptInjectionManager.java)
  at com.neeva.app.browsing.BaseBrowserWrapper.registerTabCallbacks (BaseBrowserWrapper.java:6)
  at com.neeva.app.browsing.BaseBrowserWrapper.reregisterTabIfNecessary (BaseBrowserWrapper.java:6)
  at com.neeva.app.browsing.BaseBrowserWrapper.registerBrowserCallbacks (BaseBrowserWrapper.java:8)
  at com.neeva.app.browsing.BaseBrowserWrapper.getNeevaConstants (BaseBrowserWrapper.java)
  at com.neeva.app.browsing.BaseBrowserWrapper.createAndAttachBrowser (BaseBrowserWrapper.java:10)
  at com.neeva.app.NeevaActivity.prepareBrowser (NeevaActivity.java:8)
  at com.neeva.app.NeevaActivity$onCreate$2$1$1.invokeSuspend (NeevaActivity.java:8)
  at com.neeva.app.NeevaActivity$onCreate$2$1$1.invoke (NeevaActivity.java:2)
  at kotlinx.coroutines.flow.FlowKt__MergeKt$mapLatest$1.invokeSuspend (FlowKt__MergeKt.java)
  at kotlinx.coroutines.flow.FlowKt__MergeKt$mapLatest$1.invoke (FlowKt__MergeKt.java)
  at kotlinx.coroutines.flow.internal.ChannelFlowTransformLatest$flowCollect$3$1$2.invokeSuspend (ChannelFlowTransformLatest.java:2)
[...]

PR #1004, opened 2022-09-26
Avoid some crashes

dan-alcantara-neeva:
* Add boundary check when the user is scrolling through their archived tabs.  Because the list is being paged in asynchronously, you can get into a situation where the list is wiped while the UI is trying to index into the list, resulting in an `OutOfBoundsException`.

* Catch and drop `IllegalArgumentException`s thrown by WebLayer when trying to initialize Cookie Cutter.  Play doesn't provide us the error message, so until we switch to Firebase or some other crash logging framework we can't really debug this.  Because Cookie Cutter is optional, it should be "fine" to do nothing when it can't be turned on for a particular tab.

* Figma link (if this is a UI change)
* Link to where the change is discussed (e.g. meeting notes, design docs)

### How was this tested?
- [ ] I tested this manually
- [X] I added automated tests

### Manual test cases
Because these crashes are so random, we can't really test these.

## Issues addressed
* Fixes https://github.com/neevaco/neeva-android/issues/1002
* Ignores https://github.com/neevaco/neeva-android/issues/1003


Kokobe:
Add the firebase logging as an issue TO-DO?

app/src/test/java/com/neeva/app/cardgrid/archived/ProcessedArchivedTabsTest.kt
```diff
@@ -111,6 +111,33 @@ class ProcessedArchivedTabsTest : BaseTest() {
...
+            // We should get a valid item.
+            expectThat(getTabData(4)).isEqualTo(tabDataList[2])
+
+            // Say that the PagedList has been emptied out.
+            every { lazyTabs.itemCount } returns 0
+
+            // Confirm that we going out of bounds doesn't cause a crash and returns null.
+            expectThat(getTabData(4)).isNull()
+        }
+    }
```
Kokobe, line 139: Add a test for your new boundary condition where tabs.itemCount == 0 in getTabData()

dan-alcantara-neeva:
> Add the firebase logging as an issue TO-DO?

There's a difference between TODOs and issues:
* TODOs are _supposed_ to be handled quickly as follow up.
* Issues are filed when something will take longer, or will be punted for a bit.

Logging has been on the backburner for a bit, but Chung has started looking into adding support for another framework that might tangentially give us better crash logs.

dan-alcantara-neeva:
> Add the firebase logging as an issue TO-DO?

There's a difference between TODOs and issues:
* TODOs are _supposed_ to be handled quickly as follow up.
* Issues are filed when something will take longer, or will be punted for a bit.

Logging has been on the backburner for a bit, but Chung has started looking into adding support for another framework that might tangentially give us better crash logs.

app/src/test/java/com/neeva/app/cardgrid/archived/ProcessedArchivedTabsTest.kt
dan-alcantara-neeva: Doesn't really work here; the 0th tab item would be a header for the items below it, so there's nothing to index into.

Kokobe: I'm confused, why can't you do: 
```
 val tabDataList = createTabDataList(listOf())
```

dan-alcantara-neeva: Didn't really want to, but duplicated a bit of the test above it and null checked everything.  Ideally unit tests wouldn't overlap so much, but you can't avoid it sometimes if you want to be clear about what's happening.

dan-alcantara-neeva: > I'm confused, why can't you do:
> 
> ```
>  val tabDataList = createTabDataList(listOf())
> ```

That's not what the test is checking.  The `ProcessedTabData` processes the `TabList` once and then uses that to index into the `PagedList` of `Tab`s.  The app crashes because the indices into the `PagedList` get out of sync and we still try to access the item at a particular index in the `PagedList` after the list has been cleared out.

Kokobe: Oh I get it now!

Kokobe (approved):

app/src/test/java/com/neeva/app/cardgrid/archived/ProcessedArchivedTabsTest.kt
dan-alcantara-neeva: Added some comments to the `ProcessedArchivedTabs` @Kokobe ; can you take a look and see if it helps?

Kokobe (approved):
new comment documentation looks good! Thank you for adding it

PR #1007, opened 2022-09-27
Rename Cookie Cutter to Neeva Shield

darinwf:
Uses the generic name Content Filter for the code to be more resilient to any
future such rebranding.

Add descriptive text below the global Neeva Shield toggle explaining what happens
when you disable the feature or change it.

Aligns copy to latest mocks.

Further visual polish and alignment is needed to fully conform to mocks, but that
will be a follow-up PR.

Issue #954


darinwf:
In the "Clear Browsing Data" dialog we might want to alter the text from "Neeva Shield" to something like "Neeva Shield overrides" or something suitably descriptive. Maybe there is some better copy than "overrides" that could work.

app/src/main/java/com/neeva/app/contentfilter/CookieEngineJson.kt
```diff
@@ -0,0 +1,28 @@
+// Copyright 2022 Neeva Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package com.neeva.app.contentfilter
+
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+data class CookieEngineMessage<T>(val type: String, val data: T?)
```
Kokobe, line 10: Due to a `moshi` error, it is possible to crash the app if this data class uses an `enum` for `data`'s type. ([source](https://github.com/neevaco/neeva-android/pull/964)) 

I don't think this is happening in the codebase, but it's a good thing to just double check or look out for.

app/src/main/res/values-de-rDE/strings.xml
```diff
@@ -535,14 +529,11 @@
...
 
-    <!-- [Settings][Header] Groups together settings related to JavaScript tracking on web sites. -->
-    <string name="settings_trackers">Tracker-Blockierung</string>
+    <!-- [Settings][Header] Groups together settings related to JavaScript tracking & ad blocking on web sites. -->
     <string name="blocking_strength_standard_title">Standard</string>
     <string name="blocking_strength_standard_description">Blockiert die meisten Tracker. Minimiert die Unterbrechung von Anzeigen und anderen Funktionen.</string>
     <string name="blocking_strength_strict_title">Streng</string>
     <string name="blocking_strength_strict_description">Blockiert mehr Tracker. Kann Werbung und andere Funktionen auf einigen Websites beeinträchtigen.</string>
-
-    <!-- [Settings][Header] Settings related to Ad blocking on web sites. -->
```
Kokobe, line 545: Why delete this comment?

app/src/main/java/com/neeva/app/contentfilter/CookieEngineJson.kt
darinwf: Thanks for the heads up. Note, this isn't new code. I think git just decided that too many lines changed in this file to generate a diff and instead is treating it as a new file.

PR #1020, opened 2022-10-03
String updates 2022/09/26

dan-alcantara-neeva:


dan-alcantara-neeva:
All the string IDs got renamed during https://github.com/neevaco/neeva-android/pull/1007 so this won't compile anymore; abandoning this PR

PR #1028, opened 2022-10-07
Add detailed navigation logging and session id allow list

chung-lai:
* Added `PreviewSearch`, `NavigationInbound`, `NavigationOutbound` events
  *  `PreviewSearch`: when a signed out/preview user performs a Neeva search
  *  `NavigationInbound`: navigate to a neeva domain uri
  *  `NavigationOutbound`: navigate to any other uri outside of neeva domain
* Added session id allow list to ensure we only add session id to specific events that are defined on server side

### How was this tested?
- [x] I tested this manually
- [ ] I added automated tests


app/src/main/java/com/neeva/app/browsing/TabCallbacks.kt
```diff
@@ -207,7 +211,26 @@ class TabCallbacks(
             }
 
             if (shouldRecordVisit) {
-                clientLogger?.logCounter(LogConfig.Interaction.BROWSER_PAGE_LOAD, null)
+                when {
+                    navigation.uri.isNeevaSearchUri(neevaConstants) ->
+                        if (neevaUser.isSignedOut()) {
```
Kokobe, line 216: Combine line 215 and 216 into 1 line. 

The reason is you could be using a NeevaSearchUri but are signed in -> which will result in nothing being logged.

chung-lai: yes, that's on purpose. We don't want to log an event in the case for signed in user performing a search.

app/src/main/java/com/neeva/app/logging/LogConfig.kt
```diff
@@ -34,7 +34,10 @@ class LogConfig {
         AUTH_IMPRESSION_OTHER("AuthOthersSignUpImpression", Category.FIRST_RUN),
         AUTH_IMPRESSION_SIGN_IN("AuthSignInImpression", Category.FIRST_RUN),
         AUTH_SIGN_UP_WITH_GOOGLE("AuthOptionsSignupWithGoogle", Category.FIRST_RUN),
-        AUTH_SIGN_UP_WITH_MICROSOFT("AuthOptionsSignupWithMicrosoft", Category.FIRST_RUN),
+        AUTH_SIGN_UP_WITH_MICROSOFT(
+            "AuthOptionsSignupWithMicrosoft",
+            Category.FIRST_RUN
+        ),
```
Kokobe, line 40: Any reason this can't fit in 1 line?

app/src/main/java/com/neeva/app/browsing/TabCallbacks.kt
Kokobe: Then why write lines 222 - 226?

chung-lai: oh, the user may navigate to any other neeva.com pages (like marketing, faq, help page...)

Will add comment to explain it in the code, it's definitely not obvious 😓

app/src/main/java/com/neeva/app/logging/LogConfig.kt
chung-lai: somehow it goes beyond the max length with this lol

app/src/main/java/com/neeva/app/browsing/TabCallbacks.kt
Kokobe: Oh shoot my bad too, I thought they were the same function from the naming

chung-lai: will also move it out to be it's own function to do this logging

app/src/main/java/com/neeva/app/logging/LogConfig.kt
Kokobe: LOL this is actually okay because the grey "interactionName" part doesn't actually take up real character space.

Screen Shot 2022-10-07 at 11 52 30 AM

Kokobe (approved):
Lgtm but I still think you can make the AUTH_SIGN_UP_WITH_MICROSOFT in `LogConfig.kt` one line (see my comment reply) 

If you are unsure, you can make it 1 line and then run ` ./gradlew ktlintFormat` in your terminal. This will throw an error if it is above the character limit! Let me know if you have any questions

app/src/main/java/com/neeva/app/logging/LogConfig.kt
chung-lai: Ah, makes sense, thanks!

app/src/main/java/com/neeva/app/browsing/TabCallbacks.kt
```diff
@@ -54,7 +56,9 @@ class TabCallbacks(
     private val contentFilterModel: ContentFilterModel,
     domainProvider: DomainProvider,
     private val scriptInjectionManager: ScriptInjectionManager,
-    private val clientLogger: ClientLogger?
+    private val clientLogger: ClientLogger?,
+    private val neevaConstants: NeevaConstants,
+    private val neevaUser: NeevaUser
```
dan-alcantara-neeva, line 61: @chung-lai @Kokobe This should have been made nullable like the `ClientLogger` so that we could have avoided checking it if we don't want to log anything in (say) Incognito, which explicitly tries to avoid using the `NeevaUser`.  It's not actually used unless `clientLogger != null` anyway.

Kokobe: Ah good catch, did you mean to also log incognito searches @chung-lai ?

chung-lai: thanks for catching this! no, i think we should stick to the policy which we don't log anything in incognito mode?

PR #1036, opened 2022-10-11
Add NeevaScope reddit indicator

yasminyin96:
* Add `ENABLE_NEEVASCOPE_TOOLTIP`(enabled by default) and `BLOOM_FILTER_DOWNLOAD`(disabled by default) feature flag

* Add try Reddit tooltip and blue dot: (Example website that hits bloom filter: en.wikipedia.org/wiki/bob's_burgers)
   * promo tooltip is limited to 5 times

* Add unit test for`WorkManager`

* [Figma](https://www.figma.com/file/SQpqqCk736DJWbXANK4K5w/Cheatsheet-%2F-NeevaScope?node-id=2447%3A15222)
* [Figma comments](https://www.figma.com/file/SQpqqCk736DJWbXANK4K5w?node-id=2316:20746#289184801)

### How was this tested?
- [X] I tested this manually
- [X] I added automated tests

Screen Shot 2022-10-18 at 11 24 44 AM

https://user-images.githubusercontent.com/104786010/198712801-92c17b9f-db6e-4c94-a462-8419f02d6778.mov

https://user-images.githubusercontent.com/104786010/198712456-acf99549-8cb1-4091-84b9-52034fd5b65f.mov

Try NeevaScope tooltip

https://user-images.githubusercontent.com/104786010/198712530-a7dbef5e-6d7e-46c5-bb8a-78a117caea6d.mov

https://user-images.githubusercontent.com/104786010/198712527-db913ed2-8f99-41af-9299-3ed63f2b1dd9.mov

https://user-images.githubusercontent.com/104786010/198712627-8c8449ff-9695-4c60-9839-0bda0a2333ae.mov

https://user-images.githubusercontent.com/104786010/198712637-27836ba8-a449-4df4-a881-9fbcac8f3ca6.mov


dan-alcantara-neeva:
Can you confirm with UX that this is the right flow for the first time?  It's a little weird that we tell them to open NeevaScope with a popup and then make them hit "Got It" in the intro again.

app/src/main/java/com/neeva/app/sharedprefs/SharedPrefFolder.kt
```diff
@@ -134,4 +129,22 @@ sealed class SharedPrefFolder(internal val folderName: String) {
             defaultValue = ""
         )
     }
+
+    object NeevaScope : SharedPrefFolder("NEEVASCOPE") {
```
dan-alcantara-neeva, line 133: Don't make another SharedPrefFolder.
1) Each of these SharedPreferences folders is saved as an XML file, which means that we add yet another file to read at startup.
2) This also means that we lose the old value of the SharedPref because it's in a different place.

app/src/main/java/com/neeva/app/neevascope/NeevaScopePromoModel.kt
```diff
@@ -0,0 +1,96 @@
...
+import android.net.Uri
+import com.neeva.app.sharedprefs.SharedPrefFolder
+import com.neeva.app.sharedprefs.SharedPreferencesModel
+
+enum class NeevaScopePromoType {
+    TRY_NEEVASCOPE,
+    TRY_UGC
+}
+
+class NeevaScopePromoModel(
```
dan-alcantara-neeva, line 16: Do you need another model for this?  This seems like it should live in the `NeevaScopeModel`.

app/src/main/java/com/neeva/app/browsing/toolbar/ToolbarButtons.kt
```diff
@@ -89,15 +96,31 @@ fun NeevaScopeButton(
...
-        Icon(
-            painter = painterResource(id = R.drawable.ic_neeva_logo),
-            contentDescription = stringResource(id = R.string.neevascope),
-            modifier = Modifier.size(Dimensions.SIZE_ICON_TOOLBAR),
-            tint = iconColor
-        )
+        BadgedBox(
+            badge = {
+                Badge(
+                    modifier = Modifier.size(Dimensions.NEEVASCOPE_TIP_SIZE),
```
dan-alcantara-neeva, line 108: This isn't a tip, right?  You should use an appropriately named constant, or just set this as `8.dp` because it's inherently different (a circle versus the tip of your popup).

app/src/main/java/com/neeva/app/browsing/toolbar/BrowserToolbarModel.kt
```diff
@@ -17,7 +19,12 @@ abstract class BrowserToolbarModel {
     abstract fun goForward()
     abstract fun reload()
     abstract fun reloadAfterContentFilterAllowListUpdate()
+
     abstract fun showNeevaScope()
+    abstract fun showNeevaScopeTooltip(): NeevaScopePromoType?
```
dan-alcantara-neeva, line 24: The function name doesn't make much sense: it sounds like a verb rather than a check.  Call it
`getNeevaScopePromoType` instead because it's more accurate.

app/src/main/java/com/neeva/app/browsing/IncognitoBrowserWrapper.kt
```diff
@@ -114,6 +117,12 @@ class IncognitoBrowserWrapper private constructor(
         neevaConstants = neevaConstants,
         onRemovedFromHierarchy = onRemovedFromHierarchy,
         settingsDataModel = settingsDataModel,
+        bloomFilterManager = BloomFilterManager(
```
dan-alcantara-neeva, line 120: Do you really need different `BloomFilterManager`s?  Seems like you should just create a `@Singleton` in `NeevaAppModule` so that it can be shared by both the incognito and regular profile.  The `BloomFilterManager` doesn't record any data when it tries to check the bloom filter, right?

app/src/main/java/com/neeva/app/neevascope/NeevaScopeTooltip.kt
```diff
@@ -91,33 +136,47 @@ fun Tooltip(
...
+                    anchorBounds.top - popupContentSize.height + tooltipOffset / 2
                 }
             )
         }
     }
 
     if (showTooltip.value) {
+        if (isRedditTooltip) {
+            LaunchedEffect(showTooltip) {
+                delay(4000L)
```
dan-alcantara-neeva, line 155: 1. Is this an arbitrary delay?  Where did it come from?
2. Use `TimeUnit.SECONDS.toMillis(4)` so it's obvious that you're waiting 4 seconds.

app/src/main/java/com/neeva/app/ui/theme/Dimensions.kt
```diff
@@ -25,6 +25,8 @@ object Dimensions {
     val SIZE_ICON_MEDIUM = 24.dp
     val SIZE_ICON_TOOLBAR = 21.dp
 
+    val NEEVASCOPE_TIP_SIZE = 8.dp
```
dan-alcantara-neeva, line 28: This should just live in the file with the `NeevaScopeTooltip` composable after you remove the inappropriate use case.

app/src/main/java/com/neeva/app/sharedprefs/SharedPrefFolder.kt
```diff
@@ -134,4 +129,22 @@ sealed class SharedPrefFolder(internal val folderName: String) {
...
+        )
+        val SeenNeevaScopeIntro = SharedPrefKey(
+            NeevaScope,
+            preferenceKey = "SEEN_NEEVASCOPE_INTRO",
+            defaultValue = false
+        )
+        val ShowTooltipCount = SharedPrefKey(
+            NeevaScope,
+            preferenceKey = "SHOW_TOOLTIP_COUNT",
+            defaultValue = 5
```
dan-alcantara-neeva, line 147: Is 5 the correct limit?

app/src/main/java/com/neeva/app/neevascope/NeevaScopePromoModel.kt
```diff
@@ -0,0 +1,96 @@
...
+                SharedPrefFolder.NeevaScope.ShowTooltipCount.set(
+                    sharedPreferencesModel,
+                    SharedPrefFolder.NeevaScope.ShowTooltipCount.get(sharedPreferencesModel) - 1
+                )
+            }
+        }
+        _promoCache[uri] = state
+    }
+
+    fun getNeevaScopePromoType(uri: Uri): NeevaScopePromoType? {
```
dan-alcantara-neeva, line 74: Do you have a flag that will disable this feature (and the Bloom Filter download) entirely?  You shouldn't ever develop and add a bigger feature that could cause problems without gating it behind a feature flag.  Bonus points if this enable/kill switch is controllable from the backend, but we currently don't have support for that on the Android client.

app/src/main/java/com/neeva/app/browsing/toolbar/BrowserBottomToolbar.kt
```diff
@@ -76,11 +85,31 @@ fun BrowserBottomToolbar(
                 modifier = Modifier.weight(1.0f)
             )
 
-            if (showNeevaScopeTooltip) {
-                NeevaScopeTooltip(isLandscape = false)
+            if (showNeevaScopeTooltip == NeevaScopePromoType.TRY_NEEVASCOPE) {
```
dan-alcantara-neeva, line 88: Use a `when` for enums.

app/src/main/java/com/neeva/app/browsing/toolbar/BrowserToolbar.kt
```diff
@@ -112,6 +114,15 @@ fun BrowserToolbar(
             val isDesktopUserAgentEnabled = navigationInfoFlow.value.desktopUserAgentEnabled
             val isForwardEnabled = navigationInfoFlow.value.canGoForward
 
+            val showNeevaScopeTooltip: NeevaScopePromoType? = remember {
```
dan-alcantara-neeva, line 117: How does this get updated?  AFAICT this will return a constant `NeevaScopePromoType?` that doesn't track anything.  Other places use a `Flow` and `collectAsState`.

app/src/main/java/com/neeva/app/browsing/toolbar/BrowserToolbar.kt
```diff
@@ -198,8 +210,28 @@ fun BrowserToolbar(
                     )
                 }
 
-                if (browserToolbarModel.useSingleBrowserToolbar && showNeevaScopeTooltip) {
-                    NeevaScopeTooltip(isLandscape = true)
+                if (browserToolbarModel.useSingleBrowserToolbar) {
+                    if (showNeevaScopeTooltip == NeevaScopePromoType.TRY_NEEVASCOPE) {
```
dan-alcantara-neeva, line 214: Use a `when` for enums.

app/src/main/java/com/neeva/app/neevascope/BloomFilterDownloadWorker.kt
```diff
@@ -120,8 +120,8 @@ class BloomFilterDownloadWorker(
 
     private suspend fun downloadFile() = withContext(Dispatchers.IO) {
         // Check url validity
-        assert(reddit.checksumUrl.toString().endsWith(".bin"))
-        assert(reddit.filterUrl.toString().endsWith(".json"))
+        assert(reddit.checksumUrl.toString().endsWith(".json"))
```
dan-alcantara-neeva, line 123: Any idea why this check was backwards, and why it didn't crash when it originally landed?

app/src/main/java/com/neeva/app/neevascope/NeevaScopePromoModel.kt
```diff
@@ -0,0 +1,96 @@
...
+            val hitFilter = PromoState(showRedditTooltip = true, showRedditDot = false)
+        }
+    }
+
+    enum class PromoTransition {
+        DISMISS_TOOLTIP, DISMISS_DOT
+    }
+
+    private val _promoCache: MutableMap<Uri, PromoState> = mutableMapOf()
+    val promoCache get() = _promoCache
```
dan-alcantara-neeva, line 35: This just returns another `MutableMap` because you didn't specify that it needs to be a regular `Map`, so the caller can still edit what it gets back.

app/src/main/java/com/neeva/app/neevascope/NeevaScopePromoModel.kt
```diff
@@ -0,0 +1,96 @@
...
+        if (uri == Uri.EMPTY || promoCache[uri] != null) return
+
+        val canonUrl = CanonicalUrl().canonicalizeUrl(uri, true)
+        if (canonUrl == Uri.EMPTY) {
+            _promoCache[uri] = PromoState.missFilter
+            return
+        }
+
+        val filterResult: Boolean = bloomFilterManager.contains(canonUrl.toString()) ?: return
+
```
dan-alcantara-neeva, line 47: 1. Take out the ?: part
2. Use a when to make it expllcit:

```
val filterResult: Boolean? = bloomFilterManager.contains(canonUrl.toString())
when (filterResult) {
    // The filter is still loading.
    null -> return

    true -> _promoCache[uri] = PromoState.hitFilter
    false -> _promoCache[uri] = PromoState.missFilter
}
```

dan-alcantara-neeva (changes requested):
Took a first pass...

app/src/main/java/com/neeva/app/sharedprefs/SharedPrefFolder.kt
yasminyin96: Found in cheatsheet weekly meeting note, Try neevascope promo is limited to 5 times

Screen Shot 2022-10-19 at 11 55 08 AM

app/src/main/java/com/neeva/app/neevascope/BloomFilterDownloadWorker.kt
yasminyin96: Checksum file should be ended with `.json`, when I added the test for the downloading pipeline, I changed this by mistake. I will add one more test for this `DownloadWorker` to check that task is enqueued successfully.

app/src/main/java/com/neeva/app/neevascope/NeevaScopeTooltip.kt
```diff
@@ -171,22 +237,52 @@ fun TooltipContent(
...
 }
+
+@PortraitPreviews
+@Composable
+fun TooltipRedditContent_Preview() {
+    LightDarkPreviewContainer {
+        TooltipContent(
+            horizontalPadding = Dimensions.PADDING_MEDIUM,
+            verticalPadding = Dimensions.PADDING_SMALL,
+            maxWidth = 240.dp
```
dan-alcantara-neeva, line 269: 1. Why the diff between 240.dp and 220.dp?
2. Seems like these should be constants at the top of the file because you use them both twice.

app/src/main/java/com/neeva/app/neevascope/NeevaScopeModel.kt
```diff
@@ -165,6 +197,81 @@ class NeevaScopeModel(
...
+                    SharedPrefFolder.App.ShowTooltipCount.set(
+                        sharedPreferencesModel,
+                        SharedPrefFolder.App.ShowTooltipCount.get(sharedPreferencesModel) - 1
+                    )
+                }
+
+                if (SharedPrefFolder.App.ShowTooltipCount.get(sharedPreferencesModel) <= 0) {
+                    state.showRedditTooltip = false
+                }
+            }
```
dan-alcantara-neeva, line 238: add newline

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -91,29 +97,31 @@ class BloomFilterManager(
...
-
-            /**
-             * isFinished() returns true if the State of WorkRequest is considered finished.
-             * @return True for SUCCEEDED, FAILED, and CANCELLED states
-             * Note that PeriodicWorkRequests will never enter this state (they will simply go back
-             * to ENQUEUED and be eligible to run again)
-             */
-            if (workInfo == null || workInfo.state.isFinished) {
+        if (filterDownloadEnabled) {
+            val workerRequestId = SharedPrefFolder.App.DownloadRequestId.get(sharedPreferencesModel)
```
dan-alcantara-neeva, line 101: Make another function for all of this worker/downloading logic and call it from here.  It looks like it's all fairly well encapsulated

app/src/main/java/com/neeva/app/browsing/toolbar/BrowserToolbar.kt
```diff
@@ -198,8 +213,34 @@ fun BrowserToolbar(
                     )
                 }
 
-                if (browserToolbarModel.useSingleBrowserToolbar && showNeevaScopeTooltip) {
-                    NeevaScopeTooltip(isLandscape = true)
+                if (browserToolbarModel.useSingleBrowserToolbar && isNeevaScopeTooltipEnabled) {
+                    browserToolbarModel.updateNeevaScopePromoType()
+
+                    when (neevaScopeTooltipType) {
+                        NeevaScopePromoType.TRY_NEEVASCOPE -> {
```
dan-alcantara-neeva, line 220: Is this code basically the same as the other place except that you call `isLandscape = true`?  If so, can you refactor it out and call it from both places?

dan-alcantara-neeva (changes requested):

app/src/main/java/com/neeva/app/neevascope/NeevaScopeTooltip.kt
yasminyin96: 1. Based on [Figma](https://www.figma.com/file/SQpqqCk736DJWbXANK4K5w/Cheatsheet-%2F-NeevaScope?node-id=2317%3A14205), the content size, padding of the reddit tooltip and `Try NeevaScope!` tooltip are different.

> `Try NeevaScope` only has text with 180 width and 20 horizontal padding

> Reddit tooltip has text with 174 width, reddit icon with 32 width, 10 padding between them, and 12 horizontal padding

app/src/main/java/com/neeva/app/sharedprefs/SharedPrefFolder.kt
```diff
@@ -95,6 +90,11 @@ sealed class SharedPrefFolder(internal val folderName: String) {
             preferenceKey = "REQUESTED_INSTALL_REFERRER",
             defaultValue = false
         )
+        val NeevaScopeTooltipCount = SharedPrefKey(
+            App,
+            preferenceKey = "SHOW_TOOLTIP_COUNT",
```
dan-alcantara-neeva, line 95: `SHOW_NEEVASCOPE_TOOLTIP_COUNT`

app/src/main/java/com/neeva/app/neevascope/NeevaScopeTooltip.kt
```diff
@@ -91,33 +174,47 @@ fun Tooltip(
...
                 }
             )
         }
     }
 
     if (showTooltip.value) {
+        if (isRedditTooltip) {
+            LaunchedEffect(showTooltip) {
+                delay(TimeUnit.SECONDS.toMillis(4))
+                showTooltip.value = false
```
dan-alcantara-neeva, line 194: Pull these two lines out into a lambda that gets called from here and on line 201?  Would make dismissal of the popup consistent in case you add other things that need to be done.

```
val dismissLambda = {
    showTooltip.value = false
    if (performPromoTransition != null) performPromoTransition()
}
```

app/src/main/java/com/neeva/app/browsing/toolbar/BrowserBottomToolbar.kt
```diff
@@ -36,19 +41,26 @@ fun BrowserBottomToolbar(
 ) {
     val browserWrapper = LocalBrowserWrapper.current
     val bottomOffsetDp = with(LocalDensity.current) { bottomOffset.toDp() }
+
+    val settingsDataModel = LocalSettingsDataModel.current
+    val isNeevaScopeTooltipEnabled: Boolean =
+        settingsDataModel.getToggleState(SettingsToggle.ENABLE_NEEVASCOPE_TOOLTIP).value
```
dan-alcantara-neeva, line 47: Guess you don't want to collect this State, right?  If you don't, then use instead
`getSettingsToggleValue`

app/src/main/java/com/neeva/app/browsing/BaseBrowserWrapper.kt
```diff
@@ -224,6 +229,18 @@ abstract class BaseBrowserWrapper internal constructor(
     /** Tracks configuration changes that affect the bottom toolbar. */
     private var bottomToolbarExistsJob: Job? = null
 
+    private val urlFlow: StateFlow<Uri> get() = activeTabModel.urlFlow
+    val neevaScopeTooltipTypeFlow: StateFlow<NeevaScopePromoType?> = urlFlow
+        .map {
+            if (settingsDataModel.getToggleState(SettingsToggle.ENABLE_NEEVASCOPE_TOOLTIP).value) {
```
dan-alcantara-neeva, line 235: Can avoid the return I think:

```
.map {
    if (settingsDataModel.getToggleState(SettingsToggle.ENABLE_NEEVASCOPE_TOOLTIP).value) {
        neevaScopeModel.updateUrl(it)
        neevaScopeModel.neevaScopeTooltipTypeFlow.value
    } else {
        null
    }
}
```

Otherwise, you could try:
```
return@map if (settingsDataModel.getToggleStateValue(SettingsToggle.ENABLE_NEEVASCOPE_TOOLTIP)) {
    neevaScopeModel.updateUrl(it)
    neevaScopeModel.neevaScopeTooltipTypeFlow.value
} else {
    null
}
```

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -51,13 +53,17 @@ class BloomFilterManager(
...
+    private val sharedPreferencesModel: SharedPreferencesModel,
+    private val settingsDataModel: SettingsDataModel
 ) {
     companion object {
         private const val TAG = "BloomFilterManager"
         private const val DIRECTORY_BLOOM_FILTER = "bloom_filter"
     }
 
+    val filterDownloadEnabled =
+        settingsDataModel.getToggleState(SettingsToggle.ENABLE_NEEVASCOPE_TOOLTIP).value
```
dan-alcantara-neeva, line 65: Is this reading the right shared pref?  Looks like you're downloading filters only if you're still showing the tooltips.

dan-alcantara-neeva (changes requested):

app/src/main/java/com/neeva/app/neevascope/NeevaScopeModel.kt
```diff
@@ -72,6 +102,26 @@ class NeevaScopeModel(
         }
         .flowOn(dispatchers.io)
         .stateIn(coroutineScope, SharingStarted.Eagerly, null)
+    // endregion
+
+    // region NeevaScope promo
+    private val _urlFlow = MutableStateFlow<Uri>(value = Uri.EMPTY)
+    val urlFlow: StateFlow<Uri> = _urlFlow
```
dan-alcantara-neeva, line 109: newline

app/src/main/java/com/neeva/app/neevascope/NeevaScopeModel.kt
```diff
@@ -72,6 +102,26 @@ class NeevaScopeModel(
...
+    private val _promoCache: MutableMap<Uri, RedditPromoState> = mutableMapOf()
+    val promoCache: Map<Uri, RedditPromoState> get() = _promoCache
```
dan-alcantara-neeva, line 111: newline

app/src/main/java/com/neeva/app/neevascope/NeevaScopeModel.kt
dan-alcantara-neeva: Also be consistent and use `get() = _urlFlow`?  Doesn't make a huge difference in this case because it's just exposing another field that doesn't change, but consistency is nice

app/src/test/java/com/neeva/app/neevascope/NeevaScopeModelTest.kt
```diff
@@ -86,7 +92,11 @@ class NeevaScopeModelTest : BaseTest() {
...
             coroutineScope = coroutineScopeRule.scope,
             dispatchers = dispatchers,
-            appContext = context
+            appContext = context,
+            bloomFilterManager = bloomFilterManager,
+            neevaConstants = neevaConstants,
+            sharedPreferencesModel = sharedPreferencesModel,
+            settingsDataModel = settingsDataModel
         )
     }
```
dan-alcantara-neeva, line 101: Can you add tests here for the new functions you're adding?

app/src/main/java/com/neeva/app/neevascope/NeevaScopeModel.kt
```diff
@@ -264,7 +283,9 @@ class NeevaScopeModel(
                     )
                 }
 
-                if (SharedPrefFolder.App.NeevaScopeTooltipCount.get(sharedPreferencesModel) <= 0) {
+                if (
+                    SharedPrefFolder.App.NeevaScopeTooltipCount.get(sharedPreferencesModel) <= 0
```
dan-alcantara-neeva, line 287: Did this not fit in 100 chars?

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
```diff
@@ -97,6 +98,7 @@ class BloomFilterManager(
             return
         }
 
+        workManager.enqueue(OneTimeWorkRequestBuilder<BloomFilterDownloadWorker>().build())
```
dan-alcantara-neeva, line 101: Is this supposed to be here?  Looks like you're bypassing everything and just enqueuing the download.

app/src/main/java/com/neeva/app/browsing/BaseBrowserWrapper.kt
```diff
@@ -634,6 +622,10 @@ abstract class BaseBrowserWrapper internal constructor(
                 tabCookieCutterModel.reloadUponForeground = false
             }
         }
+
+        if (settingsDataModel.getSettingsToggleValue(SettingsToggle.ENABLE_NEEVASCOPE_TOOLTIP)) {
```
dan-alcantara-neeva, line 626: Why'd you move the code to here?  This only fires when the user switches tabs and not when the URL changes, so if the user navigates somewhere else on the same tab it won't fire.

app/src/main/java/com/neeva/app/neevascope/BloomFilterManager.kt
yasminyin96: Oops....I forgot to delete this

app/src/main/java/com/neeva/app/browsing/BaseBrowserWrapper.kt
dan-alcantara-neeva: You didn't actually need to store the flow in that case: you just need the app to do something whenever the active tab's URL changed.  You can do something like this:

```
init {
    coroutineScope.launch {
        activeTabModel.urlFlow.collectLatest {
            if (settingsDataModel.getToggleState(SettingsToggle.ENABLE_NEEVASCOPE_TOOLTIP).value) {
                neevaScopeModel.updateUrl(it)
                neevaScopeModel.neevaScopeTooltipTypeFlow.value
            }
        }
    }
}
```

app/src/main/java/com/neeva/app/browsing/BaseBrowserWrapper.kt
dan-alcantara-neeva: Saw your comment above; see if the code I posted works?

dan-alcantara-neeva (approved):

Issue #1022, opened 2022-10-03
App doesn't always open external link on first attempt


dan-alcantara-neeva:
Pinged by a user on Slack:
When I open a link from my email, the first time i tap it fails to open, it only works after the second time I go back and tap a link in my email.

Been happening since at least August, confirmed by a second user on the same thread.

Kokobe:
This doesn't seem to be an issue for me anymore (I'm on 0.89.0). Do you still get this issue @dan-alcantara-neeva?



https://user-images.githubusercontent.com/20916043/197082628-cbd5c475-eea3-449e-aa23-0f9522e4f634.mp4

dan-alcantara-neeva:
Added the `monitoring` label earlier to see if we still get reports.   Were you seeing it fail before @Kokobe ?

Issue #939, opened 2022-09-01
NPE when using `NavController` from `AppNavModelImpl`


dan-alcantara-neeva:
Versions 0.79.3, 0.80.0, 0.81.0

Stack lines don't make sense, but files do:

https://play.google.com/console/u/0/developers/6544928132232754928/app/4975546713093074924/vitals/crashes/0456d25ff1a0815a76c8fd87278bd8f0/details?installedFrom=PLAY_STORE&days=30&clustering=NEW

```
Exception java.lang.NullPointerException:
  at androidx.navigation.NavController.navigate (NavController.java:3)
  at androidx.navigation.NavController.navigate (NavController.java)
  at com.neeva.app.appnav.AppNavModelImpl.show (AppNavModelImpl.java:9)
  at com.neeva.app.appnav.AppNavModelImpl.showBrowser (AppNavModelImpl.java:2)
  at com.neeva.app.browsing.BaseBrowserWrapper.getUrlBarModel (BaseBrowserWrapper.java)
  at com.neeva.app.NeevaActivity$processIntentInternal$2.invokeSuspend (NeevaActivity.java:2)
  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (BaseContinuationImpl.java)
  at kotlinx.coroutines.DispatchedTaskKt.resume (DispatchedTaskKt.java)
  at kotlinx.coroutines.DispatchedTaskKt.resumeUnconfined (DispatchedTaskKt.java:3)
  at kotlinx.coroutines.CancellableContinuationImpl.dispatchResume (CancellableContinuationImpl.java:3)
  at kotlinx.coroutines.CancelledContinuation.makeResumed (CancelledContinuation.java)
  at kotlinx.coroutines.CancellableContinuationImpl.resumeImpl$default (CancellableContinuationImpl.java:3)
  at kotlinx.coroutines.ResumeAwaitOnCompletion.invoke (ResumeAwaitOnCompletion.java)
  at kotlinx.coroutines.JobSupport.completeStateFinalization (JobSupport.java:4)
  at kotlinx.coroutines.JobSupport.tryFinalizeSimpleState (JobSupport.java:3)
  at kotlinx.coroutines.JobSupport.makeCompleting$kotlinx_coroutines_core (JobSupport.java)
  at kotlinx.coroutines.CompletableDeferredImpl.complete (CompletableDeferredImpl.java:4)
[...]

PR #1048, opened 2022-10-13
Speculatively address AppNav crashes

dan-alcantara-neeva:
* More consistently apply readiness checks before trying to load a URL in `NeevaActivity#processIntent`.
* Try to enforce that the Navigation library is used to send the browser on the main thread.
* Try to enforce that Intents are processed on the main thread instead of `Dispatchers.default`.

### How was this tested?
- [X] I tested this manually on `staging` or `release`
- [ ] I added automated tests

### Manual test cases
Fired a bunch of external VIEW intents at the app and confirmed that things seemed to work correctly.

## Issues addressed
* Speculative fix for https://github.com/neevaco/neeva-android/issues/939
* Speculative fix for https://github.com/neevaco/neeva-android/issues/1022


Kokobe (approved):
Forcing urls to load on the main thread to avoid an appnav crash, will make it hard to load it in the background in the future right? That's a future problem we'll have to figure out though.

LGTM!

dan-alcantara-neeva:
> Forcing urls to load on the main thread to avoid an appnav crash, will make it hard to load it in the background in the future right? That's a future problem we'll have to figure out though.
> 
> LGTM!

It _shouldn't_.  Loading a tab "in the background" mainly means that you don't immediately switch to it when you open a new tab.  Might have to see what WebLayer offers there.

PR #1062, opened 2022-10-18
Bump Chromium version to 106

dan-alcantara-neeva:
Pretty sure you need to bump the version number when upgrading Chromium.


Kokobe (approved):
Great catch! Updating the doc: https://paper.dropbox.com/doc/Cutting-a-new-WebLayer-release--BrScwSa8pcNj4Pu8qwLOh5KJAg-CwJYnGeODTWWJdVoJq8gz

Kokobe:
Looks like it crashes some tests, I'll look into fixing that with Darin

Issue #941, opened 2022-09-01
Update Settings UI to match current settings mocks


dan-alcantara-neeva:
Design has sent a new set of mocks to restyle all of Settings consistently.  We need to implement them:
https://www.figma.com/file/D3pbLY6ZP5YtCjTYqsGOKO/Android-Design-System?node-id=1911%3A33669

PR #1075, opened 2022-10-24
And/restyle settings

andjd:
Updated Settings styling

* Changed section heading to use a line
    * Added "heavy" and no-text variants with previews
* Changed "ic_navigate_next" icon
* Changed font weight for headings in StackedText
* Hoisted modal state to top level in ClearBrowsingPane
    * Removed bespoke CLEAR_DATA_BUTTON SettingDataType and replaced it with a regular button
    * Added ability to define the buttonAction callback on SettingsRowData
* Added "isDangerousAction" and "isSuccessMessage" to SettingsRowData(currently used to color text red and teal)
* Removed SettingsButtonRow class and replaced its use with equivalent ClickableRow class.
* Removed support section
* Updated FeedbackView to match new designs
  * Replaced `send` text button wtih materialUI send icon.  (Used official icon, slightly different from mockup)
* Built PartiallyClickableText utility for inline link in FeedbackView

* [Figma link](https://www.figma.com/file/D3pbLY6ZP5YtCjTYqsGOKO/Android-Design-System?node-id=1911%3A33669)

### How was this tested?
- [ ] I added automated tests
- [x] I checked that `@Prev
Screenshot_1667327191
Screenshot_1667327987 
Screenshot_1667328256
Screenshot_1667328311
Screenshot_1667328030
Screenshot_1667328317
Screenshot_1667328351
Screen Shot 2022-11-01 at 2 49 45 PM
Screen Shot 2022-11-01 at 2 50 12 PM

Snackbar flow:

<video width="300" src="https://user-images.githubusercontent.com/13681211/200950977-f3e8ad12-a177-48ce-b8fc-59b08a2ff396.mp4" />

## Issues addressed
#941 

https://paper.dropbox.com/doc/Mobile-design-sync-NEW--BsIJntjPO4fy_2kUK5a4Y8ibAg-UyqytDv6DOyqykAspwBaH

Audio conversation with Ria to remove the right chevron entirely for button items in settings.


dan-alcantara-neeva:
Huh.  Not sure why your CircleCI pipeline totally failed... @Kokobe/@yasminyin96  Do you remember if either of you had to do anything special on CircleCI's website before running tests?

Kokobe:
It's possible that Andrew set up the SSO incorrectly according to this: https://support.circleci.com/hc/en-us/articles/360043002793-Troubleshooting-CircleCI-Access-After-Enabling-Github-SSO

@andjd make sure you follow https://github.com/neevaco/neeva/wiki/Git-Setup 

Your SSH key settings should look something like this: 
Screen Shot 2022-10-24 at 8 28 32 PM

Kokobe:
According to that CircleCI support link, you might want to follow their steps and revoke the permissions and add the Neeva key again.

app/src/main/java/com/neeva/app/ui/widgets/ClickableRow.kt
```diff
@@ -20,15 +20,20 @@ fun ClickableRow(
...
     val backgroundColor = MaterialTheme.colorScheme.surface
     BaseRowLayout(
-        onTapRow = actionIconParams.onTapAction.takeIf { enabled },
-        onTapRowContentDescription = actionIconParams.contentDescription,
-        endComposable = if (actionIconParams.actionType != RowActionIconParams.ActionType.NONE) {
+        onTapRow = (onTapAction ?: actionIconParams?.onTapAction).takeIf { enabled },
+        onDoubleTapRow = onDoubleTapAction.takeIf { enabled },
+        onLongTap = onLongTapAction.takeIf { enabled },
+        onTapRowContentDescription = actionIconParams?.contentDescription,
+        endComposable = if (actionIconParams != null && actionIconParams.actionType != RowActionIconParams.ActionType.NONE) {
```
dan-alcantara-neeva, line 36: Maybe do this?

```
actionIconParams
    ?.takeUnless { it.actionType == RowActionIconParams.ActionType.NONE } 
    ?.let { { RowActionIconButton(it) } }
```

dan-alcantara-neeva (commented):
Code seemed fine, but the screenshots in the PR didn't seem to register and the screenshot tests didn't finish running for some reason.  Try just running `scripts/pull-new-golden-screenshots.sh` locally using the emulator made with `scripts/create-emulator.sh` and then seeing if any diffs come out

app/src/main/java/com/neeva/app/settings/clearbrowsing/ClearBrowsingPane.kt
```diff
@@ -16,7 +20,28 @@ import com.neeva.app.ui.PortraitPreviewsDark
...
+                        .associateWith { settingsController.getToggleState(it).value },
+                    timeClearingOption
+                )
+                cleared = true
+                showDialog = false
+            },
+            dismissAction = { showDialog = false }
+        )
+    }
+    SettingsPane(settingsController, ClearBrowsingPaneData(
```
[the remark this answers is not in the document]
andjd: Also fixed by ktlintFormat

app/src/main/java/com/neeva/app/ui/widgets/ClickableRow.kt
andjd: Related Unrelated comment.

I tend to like importing enum values, so you can write this like:
`?.takeUnless { it.actionType == NONE }`

What's your thought here? it requires fighting the IDE a bit which is a downside.  It feels more readable to me, although that my be influenced by it being more like swift code.

dan-alcantara-neeva: Depends on whether the enum type is clear from where it's being used... `NONE` seems a bit too generic in this case, but it still reads alright.  Will defer to you on that

dan-alcantara-neeva:
Screenshots in the PR description are still broken @andjd

andjd (commented):
The last 4 commits were pulled in from main.

It is my habit to pull the latest version of main and git merge it into my feature branch.  This just ensures no merge conflicts.  Working on my previous codebase, it would not bring in these other commits. (I think because git merge will default to fast-forward if possible).

Looks like Github has a way to do this through their UI that works fine, so 👍

Kokobe:
Hey Andrew! For the PR description, make sure to set the screenshot image height to 300 or something. Otherwise it makes the whole thing less readable

app/src/main/java/com/neeva/app/ui/FullScreenDialogTopBar.kt
```diff
@@ -27,6 +31,7 @@ fun FullScreenDialogTopBar(
     onBackPressed: () -> Unit,
     isButtonEnabled: Boolean = true,
     buttonTitle: String? = null,
+    button: Painter? = null,
```
Kokobe, line 34: Rename to `buttonPainter`

app/src/main/java/com/neeva/app/ui/FullScreenDialogTopBar.kt
```diff
@@ -50,15 +55,30 @@ fun FullScreenDialogTopBar(
...
+                            painter = button,
+                            contentDescription = buttonTitle,
+                            tint = MaterialTheme.colorScheme.onSurface
+                        )
+                    }
+                } else {
+                    if (buttonTitle != null) {
+                        TextButton(
+                            onClick = { onButtonPressed() },
+                            enabled = isButtonEnabled
```
Kokobe, line 74: Nit: I think these can fit on one line + would look cleaner

app/src/main/java/com/neeva/app/ui/FullScreenDialogTopBar.kt
```diff
@@ -50,15 +55,30 @@ fun FullScreenDialogTopBar(
...
-                ) {
-                    Text(
-                        text = buttonTitle,
-                        maxLines = 1
-                    )
+            if (onButtonPressed != null) {
+                if (button != null) {
+                    IconButton(
+                        onClick = { onButtonPressed() },
+                        enabled = isButtonEnabled
```
Kokobe, line 62: Nit: I think these can fit on one line + would look cleaner

app/src/main/java/com/neeva/app/ui/FullScreenDialogTopBar.kt
```diff
@@ -50,15 +55,30 @@ fun FullScreenDialogTopBar(
...
+                    }
+                } else {
+                    if (buttonTitle != null) {
+                        TextButton(
+                            onClick = { onButtonPressed() },
+                            enabled = isButtonEnabled
+                        ) {
+                            Text(
+                                text = buttonTitle,
+                                maxLines = 1
```
Kokobe, line 78: Nit: I think these can fit on one line + would look cleaner

app/src/main/java/com/neeva/app/ui/SectionHeader.kt
```diff
@@ -19,23 +20,30 @@ import com.neeva.app.R
...
-            modifier = Modifier.padding(
-                horizontal = Dimensions.PADDING_LARGE,
-                vertical = Dimensions.PADDING_SMALL
-            ),
-            overflow = TextOverflow.Ellipsis
+
+        Divider(
+            color = MaterialTheme.colorScheme.surfaceVariant,
+            thickness = if (heavy) 8.dp else 1.dp
+
```
Kokobe, line 33: remove new line

app/src/main/java/com/neeva/app/ui/SectionHeader.kt
```diff
@@ -19,23 +20,30 @@ import com.neeva.app.R
...
-        Text(
-            text = stringResource(stringId),
-            style = MaterialTheme.typography.titleSmall,
-            maxLines = 1,
-            modifier = Modifier.padding(
-                horizontal = Dimensions.PADDING_LARGE,
-                vertical = Dimensions.PADDING_SMALL
-            ),
-            overflow = TextOverflow.Ellipsis
+
```
Kokobe, line 29: Remove new line

app/src/main/java/com/neeva/app/ui/SectionHeader.kt
```diff
@@ -19,23 +20,30 @@ import com.neeva.app.R
...
-    Surface(
-        tonalElevation = 2.dp,
-        color = MaterialTheme.colorScheme.surface,
-        contentColor = MaterialTheme.colorScheme.onSurfaceVariant,
-        modifier = Modifier.fillMaxWidth()
+fun SectionHeader(stringId: Int? = null, heavy: Boolean = false) {
+    Column(
+        modifier = Modifier.background(
+            color = MaterialTheme.colorScheme.surface
+        ).padding(vertical = Dimensions.PADDING_SMALL)
```
Kokobe, line 27: Reformat to:
```
        modifier = Modifier
            .background(color = MaterialTheme.colorScheme.surface)
            .padding(vertical = Dimensions.PADDING_SMALL)
```

app/src/main/java/com/neeva/app/ui/SectionHeader.kt
```diff
@@ -54,3 +62,32 @@ private fun SectionHeaderPreview() {
...
+}
+
+@Preview("SectionHeader, LTR, 1x", locale = "en")
+@Preview("SectionHeader, LTR, 2x", locale = "en", fontScale = 2.0f)
+@Composable
+private fun NoTextSectionHeaderPreview() {
+    SectionHeader()
+}
+
+@Preview("SectionHeader, LTR, 1x", locale = "en")
```
Kokobe, line 89: Change name to distinguish from your HeavySectionHeader Preview.

Also make sure to not use `SectionHeader` as the name for every one of the previews. It's an important practice so when in future PRs you can distinguish the different previews and what they are supposed to do. We usually screenshot the previews rendered and put them in the PR description.

dan-alcantara-neeva:
Looks like a bunch of tests are failing because the text on the feedback pages changed.  Also one of the tests doesn't apply anymore because the Welcome Center link is going away

dan-alcantara-neeva:
Padding on the "Share screenshot" part of Feedback also seems to have gone to 0 instead of whatever it needs to be

Kokobe (commented):
Initial comments

app/src/main/java/com/neeva/app/ui/widgets/ClickableRow.kt
```diff
@@ -20,19 +21,22 @@ fun ClickableRow(
...
-            { RowActionIconButton(actionIconParams) }
-        } else {
-            null
-        },
+        onTapRow = (onTapAction ?: actionIconParams?.onTapAction).takeIf { enabled },
+        onDoubleTapRow = onDoubleTapAction.takeIf { enabled },
+        onLongTap = onLongTapAction.takeIf { enabled },
+        onTapRowContentDescription = actionIconParams?.contentDescription,
+        endComposable = actionIconParams
+            ?.takeUnless { it.actionType == NONE }
```
Kokobe, line 38: Just use `takeIf { it.actionType != NONE }` 

The double negative can be confusing for future readers.

app/src/main/java/com/neeva/app/ui/widgets/PartiallyClickableText.kt
```diff
@@ -0,0 +1,133 @@
...
+ * A Compose-style API for building a [ClickableText] element with one or more clickable spans.
+ *
+ * Non-clickable sections can be added with [Span] elements, which can take unstlyed [Text] or
+ * an [AnnotatedString] if custom styling should be added.
+ *
+ * Clickable sections can be added with [ClickableSpan] elements, which take a string and a
+ * callback lambda.  ClickableSpan can take either unstlyed [Text], which will have the linkStyle
+ * applied, or an [AnnotatedString], in which case no additional styling will be added.
+ *
+ *
```
Kokobe, line 32: remove line 31-32

app/src/main/java/com/neeva/app/ui/widgets/PartiallyClickableText.kt
```diff
@@ -0,0 +1,133 @@
...
+    builder: (ClickableStringBuilder).() -> Unit
+) {
+
+    // Applies a typical link style if linkStyle is not specified
+    ClickableStringBuilder(baseStyle.toSpanStyle(), linkStyle)
+        .apply(builder)
+        .toClickableText(modifier)
+}
+
+class ClickableStringBuilder(val baseStyle: SpanStyle, val linkStyle: SpanStyle) {
```
Kokobe, line 54: remove val for baseStyle because it is never used as a property

app/src/main/java/com/neeva/app/feedback/FeedbackView.kt
```diff
@@ -125,86 +127,79 @@ fun FeedbackView(
...
-                )
-
-                AnimatedExpandShrink(isVisible = shareUrl) {
+                Box(
+                    Modifier.padding(
+                        horizontal = Dimensions.PADDING_LARGE,
+                        vertical = Dimensions.PADDING_MEDIUM
+                    )
+                ) {
+                    PartiallyClickableText() {
```
Kokobe, line 136: () is not necessary here

Kokobe (commented):
Second round of comments

app/src/main/java/com/neeva/app/ui/widgets/ClickableRow.kt
andjd: Not to nit, but the existing formulation isn't a double negative.  `it.actionType != none` is the negative formulation of that part of the expression.

But your point that `-if` is typically a lot easier to understand than `-unless` is noted.

app/src/main/java/com/neeva/app/NeevaActivity.kt
```diff
@@ -313,10 +313,16 @@ class NeevaActivity : AppCompatActivity(), ActivityCallbacks {
 
     override fun onStart() {
         super.onStart()
+
         activityViewModel.checkForUpdates(this)
         clientLogger.logCounter(LogConfig.Interaction.APP_ENTER_FOREGROUND, null)
         updateWidgets()
-        webLayerModel.currentBrowser.reregisterActiveTabIfNecessary()
+
+        // If the user isn't in the tab switcher when the app is brought back to the foreground,
```
dan-alcantara-neeva, line 321: You should be rebasing on top on main when you upload PRs to avoid mixing up the PRs getting reviewed

app/src/main/res/values/strings.xml
```diff
@@ -92,6 +92,13 @@
 
     <!-- [General] Label for a widget that allows the user to type in a URL. -->
     <string name="url">URL</string>
+
+    <!-- [General] whitespace between words.  Used when we string multiple string resources together.  -->
+    <string name="word_separator">\u0020</string>
```
dan-alcantara-neeva, line 97: Trying to concatenate words with spaces versus full sentences will mess up translations; we shouldn't be doing this

app/src/main/res/values-en-rGB/strings.xml
```diff
@@ -676,7 +673,7 @@
     <string name="submit_feedback_help_center_link">Visit our help center!</string>
 
     <!-- [Feedback][Toggle] Lets the user decide whether or not they want to share a screenshot with their feedback. -->
-    <string name="submit_feedback_share_screenshot">Share screenshot</string>
+    <string name="submit_feedback_screenshot">Share screenshot</string>
```
dan-alcantara-neeva, line 676: Why'd you change the key name?

app/src/androidTest/java/com/neeva/app/feedback/FeedbackViewTest.kt
```diff
@@ -72,7 +75,12 @@ class FeedbackViewTest : BaseBrowserTest() {
 
             // Click on "Visit our help center".  It should open a new tab to load the Neeva URL.
             // Not super happy with this test because is actively loads the real Neeva website.
```
dan-alcantara-neeva, line 77: Can remove line 74 of the comment now; our instrumentation tests now hit up a web server running on the phone to prevent pinging the real website.

app/src/main/java/com/neeva/app/ui/FullScreenDialogTopBar.kt
```diff
@@ -50,15 +55,21 @@ fun FullScreenDialogTopBar(
...
-            if (buttonTitle != null && onButtonPressed != null) {
-                TextButton(
-                    onClick = { onButtonPressed() },
-                    enabled = isButtonEnabled
-                ) {
-                    Text(
-                        text = buttonTitle,
-                        maxLines = 1
-                    )
+            if (onButtonPressed != null) {
```
dan-alcantara-neeva, line 58: Early exit if `onButtonPressed == null`, or use a when clause?

```
when {
    onButtonPressed == null -> return@whatever_the_lambda_is

    buttonPainter != null -> {
        ...
    }

    buttonTitle != null -> {
        ...
    }
}

```

Might be a little clearer.  Otherwise, make line 67 an `else if` instead of nesting the if into it.

app/src/main/java/com/neeva/app/ui/FullScreenDialogTopBar.kt
```diff
@@ -81,3 +92,20 @@ private fun FullScreenDialogTopBarPreview() {
         )
     }
 }
+
+@Preview("FullScreenDialogTopBar, 1x font scale", locale = "en")
```
dan-alcantara-neeva, line 96: Use `@PortraitPreviews` so that you don't have to manually define all of these.  We haven't fully converted all the other @Preview callsites yet.

app/src/main/java/com/neeva/app/settings/clearbrowsing/ClearBrowsingPaneData.kt
```diff
@@ -14,22 +14,31 @@ import com.neeva.app.settings.SettingsRowData
...
     override val topAppBarTitleResId: Int = R.string.settings_clear_browsing_data
     override val shouldShowUserName: Boolean = false
     override val data = listOf(
         SettingsGroupData(
             R.string.clear_browsing_data_on_this_device,
-            timeClearingOptionToggles
-        ),
-        SettingsGroupData(
-            rows = listOf(
+            timeClearingOptionToggles + if (dataCleared) {
```
dan-alcantara-neeva, line 28: Is there a mock for this?  Seems like you should be throwing a temporary snackbar up instead of appending a row.

app/src/main/java/com/neeva/app/settings/sharedcomposables/SettingsRow.kt
```diff
@@ -75,11 +72,12 @@ fun SettingsRow(
 
     when (rowData.type) {
         SettingsRowType.BUTTON -> {
-            SettingsButtonRow(
+            ClickableRow(
                 primaryLabel = rowDataValues.primaryLabel,
                 secondaryLabel = rowDataValues.secondaryLabel,
-                onClick = onClick,
-                onDoubleClick = onDoubleClick
+                isActionDangerous = rowData.isDangerousAction,
```
dan-alcantara-neeva, line 78: Keep the word ordering consistent?  `isActionDangerous` versus `isDangerousAction`

app/src/main/java/com/neeva/app/ui/SectionHeader.kt
```diff
@@ -54,3 +60,32 @@ private fun SectionHeaderPreview() {
         SectionHeader(titleString)
     }
 }
+
+@Preview("SectionHeader_Heavy, LTR, 1x", locale = "en")
```
dan-alcantara-neeva, line 64: `@PortraitPreviews`

app/src/main/java/com/neeva/app/ui/widgets/ClickableRow.kt
```diff
@@ -20,19 +21,22 @@ fun ClickableRow(
     primaryMaxLines: Int = 1,
     secondaryMaxLines: Int = 1,
     isActionDangerous: Boolean = false,
-    actionIconParams: RowActionIconParams,
+    onTapAction: (() -> Unit)? = null,
```
dan-alcantara-neeva, line 24: Add a `kdoc` to the function explaining the priority for the tap action

app/src/main/java/com/neeva/app/ui/widgets/ClickableRow.kt
```diff
@@ -20,19 +21,22 @@ fun ClickableRow(
...
-        onTapRowContentDescription = actionIconParams.contentDescription,
-        endComposable = if (actionIconParams.actionType != RowActionIconParams.ActionType.NONE) {
-            { RowActionIconButton(actionIconParams) }
-        } else {
-            null
-        },
+        onTapRow = (onTapAction ?: actionIconParams?.onTapAction).takeIf { enabled },
+        onDoubleTapRow = onDoubleTapAction.takeIf { enabled },
+        onLongTap = onLongTapAction.takeIf { enabled },
+        onTapRowContentDescription = actionIconParams?.contentDescription,
```
dan-alcantara-neeva, line 36: Little weird that the `contentDescription` will be pulled from the `actionIconParams` even if you ignore its action.

WDYT about making another version of this composable that just has the old signature that unwraps the `RowActionIconParams`?  Would mean that you could pass `actionIconParams?.onTapAction` as `onTapAction`, explicitly pass a `contentDescription`, and explicitly pass in what the `endComposable` should be.

app/src/main/java/com/neeva/app/settings/clearbrowsing/ClearBrowsingPane.kt
```diff
@@ -16,7 +20,31 @@ import com.neeva.app.ui.PortraitPreviewsDark
 
 @Composable
 fun ClearBrowsingPane(settingsController: SettingsController, neevaConstants: NeevaConstants) {
-    SettingsPane(settingsController, ClearBrowsingPaneData(neevaConstants))
+    var showDialog by remember { mutableStateOf(false) }
+    var cleared by remember { mutableStateOf(false) }
+    if (showDialog) {
+        ClearBrowsingDialog(
```
dan-alcantara-neeva, line 26: Is there a video attached to the PR showing the altered flows?

app/src/main/java/com/neeva/app/feedback/FeedbackView.kt
```diff
@@ -82,6 +82,7 @@ fun FeedbackView(currentURLFlow: StateFlow<Uri>) {
     )
 }
 
+@SuppressLint("UnusedMaterial3ScaffoldPaddingParameter")
```
dan-alcantara-neeva, line 85: What's the reasoning behind suppressing this lint warning?  This particular one says that you're not using the padding that is being passed down by the `Scaffold`, which is necessary for avoiding content overlapping.   It looks like it's being used on line 127, though.

app/src/main/java/com/neeva/app/ui/widgets/PartiallyClickableText.kt
```diff
@@ -0,0 +1,163 @@
...
+ * of this lambda, so unlike true compose elements, [Span] and [ClickableSpan] must be direct
+ * children.
+ * @return the string data associated with the resource
+ */
+@Composable
+inline fun PartiallyClickableText(
+    // ClickableText uses BasicText, which does not pull color, so we specifically set it here.
+    baseStyle: TextStyle = LocalTextStyle.current.copy(color = LocalContentColor.current),
+
+    // Applies a sane defaults (underline and highlighted color)
```
dan-alcantara-neeva, line 50: `a sane defaults` -> `reasonable defaults`

app/src/main/java/com/neeva/app/ui/widgets/PartiallyClickableText.kt
```diff
@@ -0,0 +1,163 @@
...
+    linkStyle: SpanStyle = baseStyle.toSpanStyle().copy(
+        color = MaterialTheme.colorScheme.primary,
+        textDecoration = TextDecoration.Underline
+    ),
+    defaultSeparator: String? = stringResource(id = R.string.word_separator),
```
dan-alcantara-neeva, line 55: We shouldn't be separating individual words when building out a string or translations will get screwed up.  Strings should be properly templatized so the translator can grok the whole thing and rearrange its parts if needed.

app/src/main/java/com/neeva/app/ui/widgets/PartiallyClickableText.kt
```diff
@@ -0,0 +1,163 @@
...
+
+        ClickableText(annotated, style = LocalTextStyle.current, modifier = modifier) { offset ->
+            annotated.getStringAnnotations(offset, offset).forEach { annotation ->
+                clickActions[annotation.tag]?.invoke()
+            }
+        }
+    }
+
+    companion object {
+        const val CLICKABLE = "Clickable"
```
dan-alcantara-neeva, line 161: `private const val ANNOTATION`?

app/src/main/java/com/neeva/app/ui/widgets/PartiallyClickableText.kt
```diff
@@ -0,0 +1,163 @@
...
+            buildAnnotatedString {
+                withAnnotation(tag = tag, annotation = CLICKABLE) {
+                    append(text)
+                }
+            }
+        )
+    }
+
+    @Composable
+    fun toClickableText(modifier: Modifier) {
```
dan-alcantara-neeva, line 150: Should probably call this `build()` to match other `Builder` patterns

app/src/main/java/com/neeva/app/ui/widgets/PartiallyClickableText.kt
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * Adds a clikable span of text.
+     * Will apply the linkStyle if provided
+     * If no linkStyle was provided, typical link styling will be applied.
+     */
+    @Composable
+    @OptIn(ExperimentalTextApi::class)
+    fun ClickableSpan(text: String, separator: String? = defaultSeparator, onClick: () -> Unit) {
+        addSeparator(separator)
+        val tag = (text to clickActions.size).hashCode().toString(32)
```
dan-alcantara-neeva, line 111: Took a bit to understand what you were doing here with these tags.  Given that you're passing in a string anyway, could you just use that as the key instead of hashing your index?  Not sure why you'd append the same strings with different actions.

dan-alcantara-neeva (changes requested):

app/src/main/res/values-en-rGB/strings.xml
andjd: String-length constraints on the linter. It was either shorten the resource name or go to an awkward multi-line function call in the code. 

This style of resource naming seems a little overly wordy to me, although I understand that avoiding conflicts is an important counter consideration.

app/src/main/java/com/neeva/app/ui/FullScreenDialogTopBar.kt
andjd: Early return is not possible here, since this is an expression in an argument to a function.  Personally, I think if ... else if is cleaner here for a few reasons.

First, order matters. since buttonTitle should still be set when there's the painter because it provides context descriptions.  This is likely to be less clear when we use a when block.

Second, if we used when, we'd need an else condition, which wouldn't be helpful. (Although maybe we'd want to throw if we get a button with a click action and no painter or title, since that would be confusing ...)

app/src/main/java/com/neeva/app/settings/clearbrowsing/ClearBrowsingPaneData.kt
andjd: The change here is because the timeClearingOptionToggles and the action button are one settings group.  They were previously put in two separate SettingsGroupData objects, but that added a thin line between them not present on the mockups.

Changing the button text to the green success message matched the existing app behavior, and the new mockups do not cover this situation.

I can change it to a snackbar if that's preferable.

app/src/main/java/com/neeva/app/ui/FullScreenDialogTopBar.kt
dan-alcantara-neeva: `IllegalArgumentException`, maybe?  Kotlin `assert`s are a little odd in that they fire for both debug and release builds, but I can't imagine this check being overly painful if it's only fired when it gets composed.

app/src/main/java/com/neeva/app/settings/clearbrowsing/ClearBrowsingPaneData.kt
dan-alcantara-neeva: Hrm, could you poke UX about the right behavior here?  We're already throwing up snackbars for when the user does something, so that pattern already exists.

app/src/main/java/com/neeva/app/ui/widgets/PartiallyClickableText.kt
andjd: I was mimicing the AnnotatedString.Builder here, which uses `toAnnotatedString()` and not `build()`  But I don't have a preference here.

I should note that unlike a pure Builder, this doesn't return an object, it creates a Composable node  🤷

app/src/main/java/com/neeva/app/ui/widgets/PartiallyClickableText.kt
andjd: This is an idiot proofing step.  It just ensures correct behavior in case you have the same highlighted text in multiple places.  E.g. "click **HERE** to read our terms, to win $10000 click **HERE**"

andjd: Extracted this to a private function and added a comment.

app/src/main/java/com/neeva/app/ui/widgets/ClickableRow.kt
Kokobe: You're right haha, I should've also said my comment was a nit-pick 🙂

app/src/main/java/com/neeva/app/feedback/FeedbackView.kt
Kokobe: What is the reason behind this? I tried deleting it and the build ran fine without it 🙂

app/src/main/java/com/neeva/app/ui/SectionHeader.kt
Kokobe: Use PortraitPreviews!

Kokobe (commented):
There are a lot of comments from @dan-alcantara-neeva that are not addressed, you should act on them or reply to them (this helps with documentation and tracking code down). 

Also you can resolve my styling comments if you did them

app/src/main/java/com/neeva/app/ui/widgets/ClickableRow.kt
andjd: Fixed it and refactored the handful of uses of the old function to use the new, explicit, signature.

app/src/main/res/values/strings.xml
andjd: Removed word_separator

app/src/main/java/com/neeva/app/ui/FullScreenDialogTopBar.kt
```diff
@@ -5,29 +5,79 @@
...
+fun FullScreenDialogTopBar(
+    title: String,
+    onBackPressed: () -> Unit
+) = FullScreenDialogTopBar(
+    title = title,
+    onBackPressed = onBackPressed,
+    isButtonEnabled = false,
+    buttonTitle = null,
+    buttonPainter = null,
+    onButtonPressed = null,
```
Kokobe, line 36: Make sure to go through your PR and remove the extra commas at the end for all the other places

app/src/main/java/com/neeva/app/ui/FullScreenDialogTopBar.kt
```diff
@@ -50,34 +100,68 @@ fun FullScreenDialogTopBar(
...
+            // Runtime exceptions shouldn't be triggered if the public functions with stricter
+            // argument requirements are used.
+            when {
+                onButtonPressed == null -> {
+                    assert(buttonPainter == null && buttonTitle == null && !isButtonEnabled) {
+                        "Button info provided without an onButtonPressed callback"
+                    }
+                }
+                buttonPainter != null &&
+                    buttonTitle != null -> {
```
Kokobe, line 112: combine lines 111 and 112

app/src/main/java/com/neeva/app/ui/widgets/ClickableRow.kt
```diff
@@ -13,35 +13,39 @@ import com.neeva.app.ui.PortraitPreviews
 import com.neeva.app.ui.layouts.BaseRowLayout
 import com.neeva.app.ui.theme.Dimensions
 
+/**
+ * TODO
```
Kokobe, line 17: Remove TODO?

app/src/main/java/com/neeva/app/ui/widgets/PartiallyClickableText.kt
```diff
@@ -0,0 +1,175 @@
...
+     */
+    @Composable
+    fun Span(text: AnnotatedString, separator: String? = defaultSeparator) {
+        addSeparator(separator)
+        annotatedStringBuilder.append(text)
+    }
+
+    /**
+     * Adds a clikable span of text.
+     * Will apply the linkStyle if provided
```
Kokobe, line 104: Combine 103-104 to 1 line. Also use complete sentences in your kdocs in this file so its consistent with the codebase

app/src/main/java/com/neeva/app/ui/widgets/RowActionIcon.kt
```diff
@@ -85,6 +81,30 @@ fun RowActionIconButton(iconParams: RowActionIconParams) {
     }
 }
 
+@Composable
+fun RowActionIconButton(
```
Kokobe, line 85: Why did you recreate the same function above?

If you want to use it correctly, call the function like this:
`RowActionIconButton( RowActionIconParams(...) )`

app/src/main/res/values/strings.xml
```diff
@@ -92,6 +92,10 @@
 
     <!-- [General] Label for a widget that allows the user to type in a URL. -->
     <string name="url">URL</string>
+
+    <!-- [General] whitespace between sentences.  Used when we string multiple string resources together.  -->
+    <string name="sentence_separator">\u0020</string>
```
Kokobe, line 97: This doesn't need to be translated, so you should either remove this or put it in the `donottranslate.xml` file

Kokobe (changes requested):
Mostly style nits. Make sure to do a quick scan through your PR before sending it on to review next time 🙂

app/src/androidTest/java/com/neeva/app/feedback/FeedbackViewTest.kt
```diff
@@ -70,9 +73,12 @@ class FeedbackViewTest : BaseBrowserTest() {
             openOverflowMenuAndClickItem(R.string.feedback)
             waitForNavDestination(AppNavDestination.FEEDBACK)
 
-            // Click on "Visit our help center".  It should open a new tab to load the Neeva URL.
-            // Not super happy with this test because is actively loads the real Neeva website.
```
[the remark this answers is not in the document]
andjd: No. I deleted this comment at Dan's suggestion.

app/src/main/java/com/neeva/app/cardgrid/archived/ArchivedTabGrid.kt
```diff
@@ -104,22 +103,16 @@ fun ArchivedTabGrid(
                 secondaryLabel = stringResource(archiveAfterOption.resourceId),
                 primaryMaxLines = Int.MAX_VALUE,
                 secondaryMaxLines = Int.MAX_VALUE,
-                actionIconParams = RowActionIconParams(
-                    onTapAction = onShowArchiveAfterOptionDialog,
-                    actionType = RowActionIconParams.ActionType.NONE
-                )
+                onTapAction = onShowArchiveAfterOptionDialog,
```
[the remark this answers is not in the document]
andjd: We can discuss more broadly, but I'm generally in favor of trailing commas.  The reason for this is that adding an argument keeps diffs simpler, as you don't have an extra changed line just to add a comma.

app/src/main/java/com/neeva/app/cardgrid/archived/ArchivedTabGrid.kt
```diff
@@ -104,22 +103,16 @@ fun ArchivedTabGrid(
...
             ClickableRow(
                 primaryLabel = stringResource(R.string.archived_tabs_clear_archived),
                 primaryMaxLines = Int.MAX_VALUE,
-                isActionDangerous = true,
-                actionIconParams = RowActionIconParams(
-                    onTapAction = onDeleteAllArchivedTabs,
-                    actionType = RowActionIconParams.ActionType.NONE
-                )
+                isDangerousAction = true,
+                onTapAction = onDeleteAllArchivedTabs,
```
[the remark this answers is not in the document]
andjd: See above.

app/src/main/java/com/neeva/app/cardgrid/archived/ArchivedTabGrid.kt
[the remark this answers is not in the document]
Kokobe: Good point, @dan-alcantara-neeva WDYT?

app/src/main/res/values/strings.xml
andjd: "All languages use a space between sentences (or even words)" is one of those false assumptions about human language. Some south-east asian and Arabic scripts don't use spaces.

The DoNotTranslate list is more for branding and other words that we don't _want_ to translate.

Kokobe: Ack!

app/src/main/java/com/neeva/app/cardgrid/archived/ArchivedTabGrid.kt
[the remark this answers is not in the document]
dan-alcantara-neeva: No strong preference here.  I've wanted to keep us aligned to the Kotlin style guide, which currently encourages trailing commas at the declaration site and leaves it up to our discretion for the callsite:
https://kotlinlang.org/docs/coding-conventions.html#trailing-commas

Andrew's reasoning seems solid, though, so we can try to adhere to that.

app/src/androidTest/java/com/neeva/app/feedback/FeedbackViewTest.kt
```diff
@@ -70,9 +73,12 @@ class FeedbackViewTest : BaseBrowserTest() {
...
-            waitForNodeWithText(getString(R.string.submit_feedback_help_center_link)).performClick()
+            waitForNodeWithText(
+                getString(R.string.submit_feedback_help_center_link),
+                substring = true
+            ).performTouchInput {
+                click(position = percentOffset(x = 0.95f, y = 0.25f))
```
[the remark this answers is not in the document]
dan-alcantara-neeva: I suggested doing this because I couldn't think of a better way to do it (following a similar StackOverflow answer).  Tests will generally have to be updated whenever anything changes, anyway, so the added fragility of this doesn't seem too bad.

app/src/main/res/values/strings.xml
dan-alcantara-neeva: +1 to both points.  `donottranslate.xml` is a convenience file so that we can group all the untranslatable strings in one place instead of interspersing `translatable=false` throughout the `strings.xml` file.  Ideally, we should have most of the debugging strings in there, but it looks like two crept in (`space_enable_invite` and `settings_neevascope`).

app/src/main/java/com/neeva/app/ui/widgets/RowActionIcon.kt
andjd: I didn't refactor it entirely because that would have been beyond scope, but the way we are using RowActionIconParams is a little bit of a code smell.  

First, we were using it for things beyond the RowActionIconButton, which is problematic since it is somewhat unexpected, and couples code together that doesn't need to be.  To support this, we had added an ActionType of None so that we could use some elements of the RowActionIconParams ... when we didn't want to show an icon. I got rid of all those uses.  

Secondly, it leads to unnecessary coupling.  It is useful in the first place since it lets you pass a set of parameters down several levels in the view hierarchy, but then the elements that want to show an IconButton become coupled to that implementation since they pass the RowActionIconParams through them.  A more flexible architecture will accept a Composable function (e.g. `endComposable: @Composable () -> Unit`) instead of the `rowActionIconParams`, since then the Composable can be anything.  It also prevents the coupling we were seeing where the middle component, instead of just passing on the params, read them and applied them in unexpected ways.

Finally, it's nicer to have a signature where you supply the arguments directly instead of to a data class. It just makes the code a little simpler and more direct.  And, if you're not passing the object down through other components, you don't need it.

app/src/main/res/values/strings.xml
andjd: "All languages use a single space character to separate sentences" is one of those assumptions about text that doesn't apply to every language.  I'm pretty sure it's true for languages we're targeting now, but doesn't apply to some Arabic scripts and south-east Asian languages.

andjd: doNotTranslate applies to features we don't _want_ to localize, typically branding.  

You'll note that, ironically, there are a bunch of Hebrew translations of strings in the doNotTranslate file...

dan-alcantara-neeva: Not really ironic.  We needed those fake strings for `@Preview`s so that we could make sure that they looked fine in RTL.

app/src/main/java/com/neeva/app/ui/widgets/RowActionIcon.kt
Kokobe: Sounds good... will you refactor this in the next PR?

app/src/main/java/com/neeva/app/settings/sharedcomposables/SettingsRow.kt
```diff
@@ -93,15 +91,17 @@ fun SettingsRow(
...
-                color = MaterialTheme.colorScheme.onSurfaceVariant,
-                modifier = Modifier.padding(
-                    vertical = Dimensions.PADDING_TINY,
-                    horizontal = Dimensions.PADDING_LARGE
+            BaseRowLayout {
+                Text(
+                    rowDataValues.primaryLabel,
+                    style = MaterialTheme.typography.bodyMedium,
+                    color = if (rowData.isSuccessMessage) {
+                        MaterialTheme.colorScheme.tertiary
```
dan-alcantara-neeva, line 99: Given that we're supposed to be using ephemeral snackbars, is this conditional (and the associated `isSuccessMessage` still necessary?

app/src/main/java/com/neeva/app/ui/widgets/PartiallyClickableText.kt
```diff
@@ -0,0 +1,175 @@
...
+            }
+        }
+    }
+
+    /**
+     * Generates a unique string to use for the tag for string annotation.
+     * The current clickActions.size is hashed into the tag to avoid hash collisions when
+     * the same base string appears multiple times.
+     *
+     * @param text
```
dan-alcantara-neeva, line 165: This isn't actually a param of your function

app/src/main/java/com/neeva/app/ui/widgets/PartiallyClickableText.kt
```diff
@@ -0,0 +1,175 @@
...
+import com.neeva.app.R
+
+/**
+ * A Compose-style API for building a [ClickableText] element with one or more clickable spans.
+ *
+ * Non-clickable sections can be added with [Span] elements, which can take unstlyed [Text] or
+ * an [AnnotatedString] if custom styling should be added.
+ *
+ * Clickable sections can be added with [ClickableSpan] elements, which take a string and a
+ * callback lambda.  ClickableSpan can take either unstlyed [Text], which will have the linkStyle
```
dan-alcantara-neeva, line 31: You've misspelled `unstlyed -> unstyled` in a few places in this kdoc

app/src/androidTest/java/com/neeva/app/feedback/FeedbackViewTest.kt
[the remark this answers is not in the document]
andjd: 👍

Agree that there _should_ be a better way. But a quick search did not find one 😢

app/src/main/java/com/neeva/app/settings/sharedcomposables/SettingsRow.kt
```diff
@@ -93,15 +91,17 @@ fun SettingsRow(
...
+                    } else {
+                        MaterialTheme.colorScheme.onSurfaceVariant
+                    },
```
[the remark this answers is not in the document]
andjd: And un-needed if-else, branch too...

app/src/main/java/com/neeva/app/ui/SectionHeader.kt
```diff
@@ -4,53 +4,66 @@
...
-            style = MaterialTheme.typography.titleSmall,
-            maxLines = 1,
-            modifier = Modifier.padding(
-                horizontal = Dimensions.PADDING_LARGE,
-                vertical = Dimensions.PADDING_SMALL
-            ),
-            overflow = TextOverflow.Ellipsis
+        Divider(
+            color = MaterialTheme.colorScheme.surfaceVariant,
+            thickness = if (heavy) 8.dp else 1.dp
```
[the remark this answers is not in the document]
andjd: Switched to use DeafultDivider or HeavyDivider here

app/src/main/java/com/neeva/app/ui/FullScreenDialogTopBar.kt
dan-alcantara-neeva: @andjd Missed a comment I think

app/src/main/java/com/neeva/app/ui/widgets/RowActionIcon.kt
```diff
@@ -52,34 +48,38 @@ data class RowActionIconParams(
         SHARE,
         SHOW_PAGE_INFO
     }
-
-    fun getContentDescriptionString(context: Context): String? {
-        if (contentDescription != null) return contentDescription
-
-        return when (actionType) {
-            ActionType.REFINE -> context.getString(R.string.refine)
```
dan-alcantara-neeva, line 60: Can you make sure that the places that use these actions have their content descriptions set correctly now that you're deleting this?

dan-alcantara-neeva (commented):
One comment, otherwise seems good to go.

andjd:
Readded a few content description cues, and found two where existing behavior doesn't have a content description ... Don't know if we want that.

dan-alcantara-neeva:
> Readded a few content description cues, and found two where existing behavior doesn't have a content description ... Don't know if we want that.

Just want to make sure that we've got parity before your PR.  We have a bunch of other a11y ones we should probably look into, already...

dan-alcantara-neeva (approved):

Kokobe (approved):

PR #1078, opened 2022-10-25
Show warning when screen overlay is detected

dan-alcantara-neeva:
If we detect that another app is passing motion events through itself to us via the "Draw over other apps" permission, show the user a one-time dialog that tells them that they might be doing something dangerous and allow them the opportunity to take the permission away from the other app.  This only matters on Android versions < 12 because they mitigated it at the system level >= Android 12.

* Add a `ScreenOverlayDetectingFrameLayout` that wraps all of our Views and tries to detect when an overlay is passing `MotionEvent`s through itself.  If it detects an obscured touch, it will throw up a dialog telling them.  We only look for the OBSCURED flag instead of PARTIALLY_OBSCURED because it would negatively affect usage of the app.

* Add a `OverlayWarningDialogManager` that encapsulates the logic for detecting the overlays.

* Adds a 5 minute timeout to the instrumentation tests to try and catch when CircleCI fails to keep running our tests.

This currently doesn't do anything during First Run; we'll probably want to add support for that once the user can actually sign up for an account there.

* Figma: https://www.figma.com/file/D3pbLY6ZP5YtCjTYqsGOKO/Android-Design-System?node-id=2331%3A38054
* Tapjacking: https://developer.android.com/topic/security/risks/tapjacking

### How was this tested?
- [X] I tested this manually on `staging` or `release`
- [X] I added automated tests

### Manual test cases
Base test steps:
1. Install another app (like Twilight) that draws over other apps
2. Start the other app
3. Start our app
4. Click anywhere on the screen
5. See the dialog pop up

Dismiss test:
6. Click on the "Don't show this again" button
7. See the dialog go away
8. Click anywhere again
9. Confirm no dialog appears

Settings test:
6. Click on the "Open Settings" button
7. Get sent to Android Settings
8. Come back to the app
9. See the dialog is gone
10. Click anywhere again
11. Confirm no dialog appears

Back button test:
6. Hit system back
7. See that the dialog stays on the screen
8. Click outside of the dialog on the scrim
9. See that the dialog stays on the screen
10. Click on the "Don't show this again" button
11. See the dialog go away

Clicking outside of the scrim does nothing and you can only dismiss by clicking one of the two dialog buttons:

https://user-images.githubusercontent.com/96092902/197897597-95fcd240-ea5b-4893-93f9-59070148cd97.mp4

Going to settings and coming back doesn't show the dialog again:

https://user-images.githubusercontent.com/96092902/197897603-35d0aa8c-f387-4458-b0cb-a944fb665a97.mp4


app/src/androidTest/java/com/neeva/app/ui/widgets/ScreenOverlayDetectingFrameLayoutTest.kt
```diff
@@ -0,0 +1,134 @@
...
+            // Say that a click event was triggered that was obscured by another app.
+            activityRule.scenario.onActivity { it.performObscuredClickEvent() }
+
+            // Send the user to the Android system settings to see what apps are allowed to draw on
+            // top of other apps.
+            expectThat(activity.lifecycle.currentState).isEqualTo(Lifecycle.State.RESUMED)
+            waitForNodeWithText(getString(R.string.screen_overlay_confirm_button))
+                .assertIsDisplayed()
+                .performClick()
+            waitFor { activity.lifecycle.currentState == Lifecycle.State.CREATED }
```
Kokobe, line 105: Shouldn't we check if `SharedPrefFolder.App.CanShowScreenOverlayWarning.get(sharedPreferencesModel)` or something is now `false`?

app/src/main/res/layout/main.xml
```diff
@@ -3,13 +3,20 @@
...
+            android:id="@+id/weblayer_fragment"
+            android:layout_width="match_parent"
+            android:layout_height="match_parent"/>
+
+        <androidx.compose.ui.platform.ComposeView
+            android:id="@+id/browser_ui"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"/>
+
+    </com.neeva.app.ui.widgets.ScreenOverlayDetectingFrameLayout>
```
Kokobe, line 21: There are performance hits when our app is listening for ScreenOverlays, how did you come to the decision that it was worth it? 

(Of course this is worth it because security is important for a privacy Browser, just wanted to know how you measured the performance difference)

app/src/main/java/com/neeva/app/ui/widgets/ScreenOverlayDetectingFrameLayout.kt
```diff
@@ -0,0 +1,137 @@
...
+        return overlayWarningDialogManager.showWarningIfNecessary(
+            event = event,
+            onEventAllowed = { super.onInterceptTouchEvent(event) }
+        )
+    }
+
+    @SuppressLint("ClickableViewAccessibility")
+    override fun onTouchEvent(event: MotionEvent?): Boolean {
+        // Consume the event and avoid passing it to the child views.
+        return true
```
Kokobe, line 52: Can you make sure this doesn't break TalkBack?

app/src/main/java/com/neeva/app/ui/widgets/ScreenOverlayDetectingFrameLayout.kt
```diff
@@ -0,0 +1,137 @@
...
+     * Shows the overlay warning if we detect that the Neeva app is handling [MotionEvent]s that
+     * were ignored by apps drawing on top of ours.
+     */
+    fun showWarningIfNecessary(event: MotionEvent?, onEventAllowed: () -> Boolean): Boolean {
+        val isWindowObscured = event?.flags?.and(OBSCURED_FLAG) == OBSCURED_FLAG
+        if (!isWindowObscured) return onEventAllowed()
+
+        if (canShowWarning == null) {
+            canShowWarning = CanShowScreenOverlayWarning.get(sharedPreferencesModel)
+        }
```
Kokobe, line 77: Did you choose to lazily initialize this `canShowWarning` variable to reduce start-up time? I thought because of `strict-mode` this is going to be inevitable. 

I guess there's no real cost to making the value of this variable `lazy`, right?

app/src/main/java/com/neeva/app/ui/widgets/ScreenOverlayDetectingFrameLayout.kt
```diff
@@ -0,0 +1,137 @@
...
+
+        return when (canShowWarning) {
+            true -> {
+                // Show the warning dialog only once.
+                canShowWarning = CanShowScreenOverlayWarning.set(sharedPreferencesModel, false)
+                popupModel.showDialog { OverlayWarningDialog() }
+
+                // Intercept the touch event so that it doesn't go through to the webpage or
+                // whatever other widgets the user might try to click on.
+                true
```
Kokobe, line 87: Double check this with talkback too!

app/src/main/res/values/strings.xml
```diff
@@ -853,4 +853,16 @@
...
     <string name="download_view">View</string>
+
+    <!-- [Screen overlay warning] Title for a warning about other apps drawing on top of our app. -->
+    <string name="screen_overlay_warning_title">Screen overlay detected</string>
+
+    <!-- [Screen overlay warning] Text for a warning about other apps drawing on top of our app. -->
+    <string name="screen_overlay_warning_text">Another app is appearing on top of your screen. It may interfere with your usage of Neeva. You may change the \"Draw over other apps\" permission in Settings. </string>
+
+    <!-- [Screen overlay warning] Opens the Settings page. -->
+    <string name="screen_overlay_confirm_button">Open Settings</string>
```
Kokobe, line 864: Nit: Maybe the comment should say it opens Android Settings to distinguish Neeva Settings.

app/src/test/java/com/neeva/app/ui/widgets/OverlayWarningDialogManagerTest.kt
```diff
@@ -0,0 +1,90 @@
...
+        CanShowScreenOverlayWarning.set(sharedPreferencesModel, false)
+
+        val event = MotionEventBuilder.newBuilder()
+            .setFlags(MotionEvent.FLAG_WINDOW_IS_OBSCURED)
+            .build()
+        expectThat(
+            overlayWarningDialogManager.showWarningIfNecessary(event, onEventAllowed)
+        ).isFalse()
+
+        verify(exactly = 1) { onEventAllowed.invoke() }
```
Kokobe, line 86: Should you verify this is `true`?

app/src/androidTest/java/com/neeva/app/ui/widgets/ScreenOverlayDetectingFrameLayoutTest.kt
dan-alcantara-neeva: Should be accounted for in the unit tests

app/src/main/res/layout/main.xml
dan-alcantara-neeva: I still haven't entirely come to the decision that it's worth it, which is why I've been sitting on this bug all week :)  Most other apps (including Chrome) totally ignore this issue, but apparently the Play Store app at least makes sure that there are no obscuring views to avoid users clicking on buttons to buy apps.

I've tried to minimize the impact to perf by putting the obscured touch check first, and only hitting SharedPreferences up if we need to.  Also considered changing the layout of the app so that we don't put the `FrameLayout` in at all if the user has already been shown the dialog once, but you're still going to be in a situation where you need to keep it in the view hierarchy until the user has had an opportunity to dismiss it.

WDYT?  Is this issue worth doing the cursory checks?

app/src/main/java/com/neeva/app/ui/widgets/ScreenOverlayDetectingFrameLayout.kt
dan-alcantara-neeva: This was meant to avoid the performance hit of checking SharedPreferences for every single motion event.  Though tbh, because SharedPreferences are cached in-memory after the first read, there's really no huge benefit for caching it again here.  Debating whether to remove it.

app/src/main/java/com/neeva/app/ui/widgets/ScreenOverlayDetectingFrameLayout.kt
dan-alcantara-neeva: Talkback works fine.

app/src/main/java/com/neeva/app/ui/widgets/ScreenOverlayDetectingFrameLayout.kt
dan-alcantara-neeva: Same as above

dan-alcantara-neeva: Clarifying: returning true from `onInterceptTouchEvent` means that you want Android to send the touch event to this View's `onTouchEvent`, so it's the same thing

dan-alcantara-neeva:
After offline discussion, we're closing this one without merging it in.

PR #1082, opened 2022-10-26
Update readme instructions for setting ANDROID_HOME

dan-alcantara-neeva:
Update the commands in the README so that it'll properly set the PATHs for the Android SDKs


Kokobe (approved):
It's really weird because I feel like it is equivalent

dan-alcantara-neeva:
> It's really weird because I feel like it is equivalent

The catch is that until you've run `source ~/.zshrc`, `ANDROID_HOME` won't be set.  That means the lines above will incorrectly use "" as the value.

PR #1092, opened 2022-11-02
Fix NeevaScope onboarding flow

yasminyin96:
* Fix NeevaScope education screen:
     * change to full-screen sheet
     * add `X` button to dismiss the sheet
     * use `Let's try it!` for button on non-Neeva page and non SRP

* Both `Try NeevaScope` tooltip and reddit one will disappear after 4 seconds or by tapping

* Fix onboarding flow:
     * The tooltip is disabled by default
     * FLOW1: Neeva/SRP -> `NeevaScope` button -> `Got it!` button -> enable tooltip(if no other promo shows, it will show NeevaScope tooltip) -> see NeevaScope content or tooltip shown 5 times -> finish onboarding and disable tooltip
     * FLOW2: non Neeva/SRP -> `NeevaScope` button -> `Let's try it` button -> NeevaScope content -> finish onboarding and disable tooltip
     * FLOW3: non Neeva/SRP -> `NeevaScope` button ->  `X` button or swipe down -> enable tooltip(if no other promo shows, it will show NeevaScope tooltip) -> see NeevaScope content or tooltip shown 5 times -> finish onboarding and disable tooltip

* NeevaScope tooltip will show ONLY when no other promo has been shown during current session and either:
     * Content filtering/Neeva Shield is disabled
     * The Content filtering/Neeva Shield promo has already been shown

* [Figma](https://www.figma.com/file/SQpqqCk736DJWbXANK4K5w/Cheatsheet-%2F-NeevaScope?node-id=2316%3A20746)
* [doc](https://paper.dropbox.com/doc/NeevaScope-Android-Tooltip--BsLfKYTH_gb5l6kY0HIe4XAeAQ-lrAFSOOoQmeXCAtu3caUf)
### How was this tested?
- [ ] I added automated tests
- [X] Existing automated tests cover all of my changes

https://user-images.githubusercontent.com/104786010/200669899-3d362377-5ca2-4368-8bd0-07c44540af2f.mov

https://user-images.githubusercontent.com/104786010/200669911-2212337f-fd25-4209-a9d9-e371abaee18a.mov

https://user-images.githubusercontent.com/104786010/199609463-109464ab-a7cc-4567-9c3c-3ea375be1855.mov

https://user-images.githubusercontent.com/104786010/199609612-5af281f6-6ede-4e54-bdd2-24216f17a6b5.mov

Screen Shot 2022-11-02 at 3 02 17 PM


app/src/main/java/com/neeva/app/browsing/toolbar/BrowserBottomToolbar.kt
```diff
@@ -61,6 +64,10 @@ fun BrowserBottomToolbar(
     val browserToolbarModel = LocalBrowserToolbarModel.current
     val showRedditDot = remember { mutableStateOf(false) }
 
+    val sharedPreference = LocalSharedPreferencesModel.current
```
dan-alcantara-neeva, line 67: "SharedPreferences" is a specific thing; call this `sharedPreferencesModel` to make it clearer

app/src/main/java/com/neeva/app/neevascope/NeevaScopeInfoScreen.kt
```diff
@@ -152,18 +166,19 @@ fun NeevaScopeInfoBody() {
 
 @Composable
 fun NeevaScopeInfoButton(
-    onTapAction: () -> Unit,
+    @StringRes buttonText: Int,
```
Kokobe, line 169: Style Nit: I think naming this `buttonTextId` helps differentiate it from other "buttonText" parameters that are actually strings instead of ints.

app/src/main/java/com/neeva/app/neevascope/NeevaScopeInfoScreen.kt
```diff
@@ -152,18 +166,19 @@ fun NeevaScopeInfoBody() {
...
+    tapButton: () -> Unit,
     modifier: Modifier
 ) {
     Button(
-        onClick = onTapAction,
+        onClick = tapButton,
```
Kokobe, line 174: Style nit: Keep the parameter names consistent: instead of `tapButton` use `onClick`

Kokobe (changes requested):
Just some style nits

app/src/main/java/com/neeva/app/browsing/toolbar/BrowserBottomToolbar.kt
```diff
@@ -95,13 +84,14 @@ fun BrowserBottomToolbar(
...
 
-            if (isNeevaScopeTooltipEnabled && didShowAdBlockOnboarding.value) {
+            if (browserToolbarModel.shouldShowNeevaScopeTooltip()) {
                 browserToolbarModel.getNeevaScopeModel()?.let {
                     NeevaScopeTooltip(
                         neevaScopeModel = it,
                         showRedditDot = showRedditDot,
                         isLandscape = false
                     )
+                    SharedPrefFolder.App.ShowingNeevaScopeTooltip.set(sharedPreferencesModel, true)
```
Kokobe, line 94: Are you sure `ShowingNeevaScopeTooltip` is a UI state we should be storing in `SharedPreferences`? 🙂 It isn't something that needs to be persisted after the app dies.

If anything it should be a `StateFlow` in `NeevaScopeModel` where the State is named `ShowNeevaScopeTooltip`. 

This condition in PopupModel.kt:
```
// val canShowPromo:
!SharedPrefFolder.App.ShowingNeevaScopeTooltip.get(sharedPreferencesModel) &&
```
can be moved upwards to `BrowserToolbarModelImpl` and changed to:

```
getNeevaScopeModel().ShowNeevaScopeTooltip.value == true
```

Kokobe: It actually could be just a variable because it looks like no compose UI is going to call `showNeevaScopeTooltip.value`. 

You can decide how this variable can be updated, I recommend piping it through `BrowserToolbarImpl` since it already has a reference to `NeevaScopeModel` but you might know a better way?

app/src/main/java/com/neeva/app/ui/PopupModel.kt
```diff
@@ -79,6 +82,11 @@ class PopupModel(
     val bottomSheetDialogState: StateFlow<BottomSheetDialogHostState?>
         get() = _bottomSheetDialogState
 
+    // Check whether any type of promo has been shown during current session
+    val canShowPromo: Boolean = bottomSheetDialogState.value == null &&
```
Kokobe, line 86: I'm pretty sure this needs to be a `get()` variable or a function. You want the value to be updated whenever some other code calls `canShowPromo`. Otherwise it's just a static value that only gets initialized during the Hilt init

Kokobe (changes requested):
Lmk if you need any help

app/src/main/java/com/neeva/app/ui/PopupModel.kt
yasminyin96: What if we only want to update it when app starts? `canShowPromo` is to check whether a promo can be shown during the current session(from the app starts to the app dies). Only when no promo has been shown during the session, `canShowPromo` will be true. (ref: [doc](https://paper.dropbox.com/doc/NeevaScope-Android-Tooltip--BsvCxZmT1xTL3iyb0SrPSijfAg-lrAFSOOoQmeXCAtu3caUf#:uid=486812739765822488961975&h2=Implementation-details))

Kokobe: Ah got it! Yeah this is good then

dan-alcantara-neeva:
Was this one good to review again @yasminyin96?

yasminyin96:
> Was this one good to review again @yasminyin96?

Yes

app/src/main/java/com/neeva/app/browsing/BaseBrowserWrapper.kt
```diff
@@ -1113,26 +1111,59 @@ abstract class BaseBrowserWrapper internal constructor(
...
         if (currentUrl != Uri.EMPTY) {
             neevaScopeModel.updateQuery(currentUrl.toString(), activeTabModel.titleFlow.value)
         }
 
-        popupModel.showBottomSheet { onDismiss ->
+        val isOnNeevaSearch =
+            activeTabModel.displayedInfoFlow.value.mode != ActiveTabModel.DisplayMode.URL
+
+        val seenNeevaScopeIntroFlow =
+            MutableStateFlow(SharedPrefFolder.App.SeenNeevaScopeIntro.get(sharedPreferencesModel))
```
dan-alcantara-neeva, line 1122: Having a little bit of trouble parsing this.  It looks like you're trying to create a `MutableStateFlow` and initializing it with the current value of the `SeenNeevaScopeIntro` `SharedPreference`, then manually editing it and keeping the `SharedPreference` in sync.

I think you want to do this instead:

```
val seenNeevaScopeIntro by SeenNeevaScopeIntro.getFlow(sharedPreferencesModel).collectAsState()
```

That way you're collecting the Flow directly.  On line 1146 where you're setting both the value of the `MutableStateFlow` and the `SeenNeevaScopeIntro` `SharedPreference`, you could just directly set the `SharedPreference` and the `Flow` should automatically update.

app/src/main/java/com/neeva/app/browsing/toolbar/BrowserBottomToolbar.kt
```diff
@@ -42,22 +41,18 @@ fun BrowserBottomToolbar(
...
     )
 }
 
 @Composable
 fun BrowserBottomToolbar(
     isIncognito: Boolean,
-    isNeevaScopeTooltipEnabled: Boolean = false,
     modifier: Modifier = Modifier
 ) {
+    val sharedPreferencesModel = LocalSharedPreferencesModel.current
```
dan-alcantara-neeva, line 55: Is this used anywhere?

yasminyin96: Not used, I will delete this line as well as the line in BrowserToolbar.

app/src/main/java/com/neeva/app/browsing/BaseBrowserWrapper.kt
```diff
@@ -1112,27 +1110,67 @@ abstract class BaseBrowserWrapper internal constructor(
 
     // region: NeevaScope
     override fun showNeevaScope() {
-        val currentUrl = urlFlow.value
-        val isOnNeevaSearch =
-            activeTabModel.displayedInfoFlow.value.mode != ActiveTabModel.DisplayMode.URL
+        if (settingsDataModel.getSettingsToggleValue(SettingsToggle.ENABLE_NEEVASCOPE_TOOLTIP)) {
+            SharedPrefFolder.App.SeenNeevaScopeIntro.set(sharedPreferencesModel, true)
+        }
```
yasminyin96, line 1115: I added these lines(1113-1115) here because if the user clicks on the `X` button instead of `Let' try it` button on non-Neeva page, we don't want to update the `SeenNeevaScopeIntro` immediately. If `SeenNeevaScopeIntro` immediately becomes true, the sheet will direct to the loading and result screen before the sheet dismissing.

app/src/main/java/com/neeva/app/ui/PopupModel.kt
```diff
@@ -79,6 +82,10 @@ class PopupModel(
...
+        SharedPrefFolder.FirstRun.didShowAdBlockOnboarding.get(sharedPreferencesModel)
```
yasminyin96, line 87: @dan-alcantara-neeva I found that `didShowAdBlockOnboarding` will be true by default and set to false for the first run. Is it possible to get the value of `didShowAdBlockOnboarding` after the first run when the app starts?

dan-alcantara-neeva: You'll want to sync with @chung-lai about that.  I think it's because of how the code doesn't set the value when it gets checked, so he had to be creative to avoid showing existing users the ad block promo.

app/src/main/java/com/neeva/app/browsing/BaseBrowserWrapper.kt
yasminyin96: Got it.

app/src/main/java/com/neeva/app/ui/PopupModel.kt
dan-alcantara-neeva: If that's the case, is your question resolved, or did you still need to work around that?

yasminyin96: Yes, it's resolved, and pushed the new commit with comment about this.

app/src/main/java/com/neeva/app/neevascope/NeevaScopeModel.kt
```diff
@@ -302,6 +307,93 @@ class NeevaScopeModel(
...
+        popupModel: PopupModel
+    ) {
+        // If user clicks on the X button instead of "Let' try it" button on non-Neeva page,
+        // do not update the SeenNeevaScopeIntro immediately.
+        if (settingsDataModel.getSettingsToggleValue(SettingsToggle.ENABLE_NEEVASCOPE_TOOLTIP)) {
+            SharedPrefFolder.App.SeenNeevaScopeIntro.set(sharedPreferencesModel, true)
+        }
+
+        if (currentUrl != Uri.EMPTY) {
+            updateQuery(currentUrl.toString(), activeTabModel.titleFlow.value)
```
dan-alcantara-neeva, line 324: Instead of passing the whole model, just pass in the `titleFlow.value` and whether the user `isOnNeevaSearch`

app/src/main/java/com/neeva/app/ui/PopupModel.kt
dan-alcantara-neeva: While you're there, can you rename that `val didShowAdBlockOnboarding = SharedPrefKey(` to be properly capitalized like every other key?  `val DidShowAdBlockOnboarding = SharedPrefKey(`

dan-alcantara-neeva (approved):

Kokobe (approved):

PR #1097, opened 2022-11-09
Add boiler-plate Billing code behind a flag

Kokobe:
In order to test out subscriptions end-to-end on the Play Console, the Play Store API wants us to first upload a build that uses the `Billing` Library. 

Screen Shot 2022-11-08 at 3 56 26 PM

I've integrated some boilerplate code from the [Google Play Billing Docs](https://developer.android.com/google/play/billing/integrate) but it doesn't do anything because it's behind a Settings Toggle flag (`DEBUG_ENABLE_BILLING`).

Planning to fully implement this in the next few PRs once I get Billing server feedback. 

* Not necessary for this PR but here's the [Android In-App Subscriptions doc](https://paper.dropbox.com/doc/Android-In-App-Subscriptions--Bskn6AsAeeKYbc5CJ~tawQEDAg-OLRpYDxe9lW7lDxwjf2Wm)
* Click-up: https://app.clickup.com/t/36034046/GR-2551

### How was this tested?
- [ ] I added automated tests

### Manual test cases
Opened app and saw that it did not crash.

Turned on the feature flag and saw that it did not crash (see recording).

### Screenrecordings

https://user-images.githubusercontent.com/20916043/200704349-de953b05-46c6-4968-9fa3-7375086d756d.mov


app/src/main/java/com/neeva/app/billing/NeevaBillingClient.kt
```diff
@@ -0,0 +1,114 @@
...
+import com.neeva.app.Dispatchers
+import com.neeva.app.settings.SettingsToggle
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+
+/** This class is currently incomplete! It is sandboxed by [SettingsToggle.DEBUG_ENABLE_BILLING]. */
+class NeevaBillingClient(appContext: Context, val dispatchers: Dispatchers) {
+    private val purchasesUpdatedListener =
+        PurchasesUpdatedListener { billingResult, purchases ->
+            // To be implemented in a later section.
```
yozuysal, line 28: let's get rid of this for now. todo should be enough

app/src/main/java/com/neeva/app/billing/NeevaBillingClient.kt
```diff
@@ -0,0 +1,114 @@
...
+                // TODO(kobec): process billingResult codes
+                when (billingResult.responseCode) {
+                    BillingResponseCode.OK -> {
+                        _productDetailsListFlow.value = productDetailsList
+                    }
+                    // TODO(kobec): handle other error codes better
+                    else -> {
+                        val errorMessage = "BillingClient ERROR: responseCode = " +
+                            "${billingResult.responseCode} " + "${billingResult.debugMessage}"
+                        // TODO(kobec): decide whether to show a popup or something...
```
yozuysal, line 67: I dont think we need this. There wont be a need for the user to do anything here, so we will most likely just handle errors with the above.

app/src/main/java/com/neeva/app/NeevaActivity.kt
```diff
@@ -261,6 +263,15 @@ class NeevaActivity : AppCompatActivity(), ActivityCallbacks {
             }
         }
 
+        // TODO(kobec): remove when Billing is ready
+        if (settingsDataModel.getSettingsToggleValue(SettingsToggle.DEBUG_ENABLE_BILLING)) {
+            lifecycleScope.launch {
+                repeatOnLifecycle(Lifecycle.State.RESUMED) {
```
yozuysal, line 269: why resetup at every resume? should we not do this one per lifecycle?

app/src/main/java/com/neeva/app/billing/NeevaBillingClient.kt
Kokobe: This handling will require a bit more thinking because in order to `launchBillingFlow()` (which is Google Pay's bottom sheet popup that enables Google Pay), we need the `productDetailsList`. 

Let's say the user suddenly doesn't have a WiFi connection. When the user tries to purchase the subscription, we won't be able to launch this flow until the `BillingClient` is setup properly and gives us this `productDetailsList`. 

When we eventually design a "Pay now" button, it's ambiguous what the behavior should be here. Should the "Pay now" button be disabled until the proper setup is ready? Should the user know (via snackbar) that they need Wifi to continue? I think both are important to look out for later

app/src/main/java/com/neeva/app/NeevaActivity.kt
Kokobe: Ah good call-out, I should've documented this in the PR description. The reason is:
1. Fetching Purchases documentation says I should and has their own reasons for it: https://developer.android.com/google/play/billing/integrate#fetch
2. It makes sense that every time you `onResume` in the app, you should restart your connection to the `BillingClient` to make sure its fresh and to check if there are any new updates to your subscription.

app/src/main/java/com/neeva/app/billing/NeevaBillingClient.kt
yozuysal: yeah, I think the above todo covers these considerations. "popup or something" is somewhat misleading imo

app/src/main/java/com/neeva/app/NeevaActivity.kt
yozuysal: ack!

yozuysal (approved):
lgtm % the final nit. I think you will have enough todo's peppered in here that they wont be missed :)

PR #1103, opened 2022-11-11
Billing: Set up launch billing flow prototype (behind a flag)

Kokobe:
## IMPORTANT: Testing

In order to test the Billing Flow out, you need to:

1. Upload a build to the Internal Track and give its name a suffix of `PAYMENTS_TESTING`
2. Download the app on your physical phone from the Google Play Store
3. Go to developer settings and enable billing flow. 
4. If connecting to the back-end, change the neeva app url to the local server you are connecting to. (Also in developer settings)

**Why?** 
The Billing Flow requires that the `applicationId` of the build you are running has to be the same as the **release** build's `applicationId`. Also it needs to be a `signed` `release` build downloaded from the Google Play Store. 

## Billing Flow
For a documented flow on how the BackEnd, Client, and Google Play Library interact in this flow, see:
* The [flow chart](https://www.figma.com/file/v2g41XtVbKj0QWmPsq4luw/Design-critique---Android-Premium-Functional-Design-Flows?node-id=101%3A1715&t=sIR6kAb0tzDt0Woi-0)
* Purchase Flow written doc: https://paper.dropbox.com/doc/Android-In-App-Subscriptions--BtJ908IofE4p2iGwMlAcgJHgAg-OLRpYDxe9lW7lDxwjf2Wm#:h2=Making-a-Purchase-Flow-(source
* [Handling Purchase Flows Errors](https://paper.dropbox.com/doc/Handling-BillingClient-Request-Errors-k4pCU6N0trZJfvFgxSWn9) 

## Description
Uses the debug `DEBUG_ENABLE_BILLING` flag to decide whether to show the `BILLING_FLOW` screen.  

* Implements the purchase flows!
* Adds `InitializeGooglePlaySubscription` to graphql and uses it to generate a `obfuscatedUserID`
* Implements error handling for failed `BillingClient` queries. See `BillingClientWrapper`

Misc: 
* Creates a `NetworkHandler` class to check if the device connected to the internet. 

### How was this tested?
- [x] I added automated tests

### Manual test cases
See screen recordings:

#### Launching the billing flow: 
https://user-images.githubusercontent.com/20916043/203181695-baa1b5a0-7b3d-4b96-b692-ccb003728dd2.mp4

#### Launching the billing flow with no internet:
https://user-images.githubusercontent.com/20916043/203181603-1f4dfffa-3cdc-4842-bd16-2711994650ad.mp4

#### Cancelling a subscription (for now):

https://user-images.githubusercontent.com/20916043/203182603-267af6a6-e062-4d1f-b3b1-a6db1ce65b38.mp4


dan-alcantara-neeva (changes requested):
Going to need to reconsider how this should work, I think.  Especially because the call to purchase a subscription is going to depend on these other two calls working

app/src/main/java/com/neeva/app/NeevaActivity.kt
```diff
@@ -263,11 +264,15 @@ class NeevaActivity : AppCompatActivity(), ActivityCallbacks {
             }
         }
 
+        Log.e("KOBE", "onCreate")
         // TODO(kobec): remove when Billing is ready
         if (settingsDataModel.getSettingsToggleValue(SettingsToggle.DEBUG_ENABLE_BILLING)) {
```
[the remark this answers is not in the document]
Kokobe: That's okay, I'll add a popup to tell you to restart the app when using this toggle.

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    fun terminateBillingConnection() {
+        billingClientWrapper.terminateBillingConnection()
+    }
+
+    fun onResume() {
+        // The Google Play Billing Integration docs say to do this every onResume:
+        // https://developer.android.com/google/play/billing/integrate#fetch
+        onBillingClientReady()
```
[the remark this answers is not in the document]
Kokobe: Changed the name to `fetchProductsAndPurchases()`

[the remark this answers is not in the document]
Kokobe: Maybe I can make the `startBillingConnection()` to be one of the jobs in the job queue. That way, fetchProductsAndPurchases will wait for the connection to be valid before doing its fetch tasks

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
```diff
@@ -0,0 +1,163 @@
...
+    val coroutineScope: CoroutineScope,
+    val dispatchers: Dispatchers,
+    val settingsDataModel: SettingsDataModel
+) {
+    private data class BillingClientJob(
+        val uniqueTaskName: String,
+        val job: suspend () -> Unit
+    )
+
+    private val jobQueue: Queue<BillingClientJob> = LinkedList()
```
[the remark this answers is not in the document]
Kokobe: Hmmmm I'm using the `private val isProcessingJob = AtomicBoolean(false)` line to manually avoid that exception. Is there another way you suggest on dealing with this?

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    /** Processes one [BillingClientJob] job at a time until the [jobQueue] is empty. */
+    private fun processJobs() {
+        if (jobQueue.isNotEmpty() && !isProcessingJob.getAndSet(true)) {
+            val billingClientJob = jobQueue.peek()
+            val job = coroutineScope.launch(dispatchers.io) {
+                billingClientJob.job()
+            }
+            job.invokeOnCompletion {
```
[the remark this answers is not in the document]
Kokobe: Very cool!!!

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientWrapper.kt
```diff
@@ -0,0 +1,94 @@
...
+        // a future use of BillingClient (i.e. fetchPurchases or fetchProductDetails) will have an
+        // error. This will trigger the old, invalid billingClient to be destroyed and replaced by a
+        // new one.
+        billingClient?.endConnection()
+    }
+
+    suspend fun retryConnection(): Boolean {
+        return suspendCoroutine { continuation ->
+            onBillingClientReadyContinuation = continuation
+            coroutineScope.launch(dispatchers.io) {
```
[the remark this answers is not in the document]
Kokobe: So what suspendCoroutine actually puts you in is a: `crossinline block: (Continuation<T>) -> Unit`

Since this isn't a `suspend` parameter, I can't add a `delay()` unless I launch another coroutine inside of it.

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    fun onResume() {
+        // The Google Play Billing Integration docs say to do this every onResume:
+        // https://developer.android.com/google/play/billing/integrate#fetch
+        onBillingClientReady()
+    }
+
+    private fun onBillingClientReady() {
+        queueJobWithRetry(
```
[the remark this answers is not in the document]
Kokobe: If one of the tasks fails, our behavior should be to not allow the user to purchase a subscription using Google Play (aka. never show the purchase screen -- which will be done in a future PR)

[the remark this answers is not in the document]
Kokobe: I've made it so that `onBillingSetupFinished()` does not call `fetchProductsAndPurchases()`. This avoids calls triggering a requeue of a call that already happened. 

Now the flow from `onCreate` is:

```
Queue = [
fetchPurchases(),
fetchProductDetails()
]

Whenever one of these operations fail, the billingClient will be restarted up to a max of 3 times 
with exponential backoff. 

The first time is guaranteed to fail, which will set up the billingClient and then re-run the operation.
```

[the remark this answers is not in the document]
Kokobe: `fetchPurchases` and `fetchProductDetails` will only be requeued in `onResume()`. If the task already exists in the job queue, it will not be queued.

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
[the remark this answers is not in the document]
Kokobe: I've enforced the queue behavior by doing a `private val jobQueueFlow: MutableSharedFlow<Queue<BillingClientJob>> = MutableStateFlow(jobQueue)`

Is that right?

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientWrapper.kt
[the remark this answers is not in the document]
dan-alcantara-neeva: Why don't you just put the delay outside of that `suspendCoroutine` block, then?  You don't care when it fires -- only that it waits 1 second before doing it.  `retryConnection` is already a suspend function, too.

app/src/main/java/com/neeva/app/billing/BillingScreen.kt
```diff
@@ -0,0 +1,106 @@
...
+    Timber.tag("KOBE").e("obfuscatedUserID = $obfuscatedUserID")
+    Timber.tag("KOBE").e("purchases = $purchases")
+    Timber.tag("KOBE").e("product details = $productDetails")
+
+    Surface {
+        Box(Modifier.fillMaxSize()) {
+            Column(Modifier.fillMaxSize()) {
+                val subscription = subscriptionManager.let {
+                    when {
+                        it.hasMonthlyPremiumPlan.collectAsState(false).value -> "monthly premium"
```
dan-alcantara-neeva, line 42: 1. You're using `collectAsState` weirdly here; you should be collecting them as state above line 37 and checking their values in your `when` block.
2. Are these string constants or just your debugging output?

app/src/main/java/com/neeva/app/billing/BillingScreen.kt
```diff
@@ -0,0 +1,106 @@
...
+                    // Monthly
+                    Button(onClick = {
+                        subscriptionManager.buy(
+                            productDetails = productDetails,
+                            purchases = purchases,
+                            tag = MONTHLY_PREMIUM_PLAN,
+                            activity = activity
+                        )
+                    }) {
+                        Text("Monthly Plan")
```
dan-alcantara-neeva, line 77: Is this real UI or just debugging?  You should be putting strings in the `strings.xml` file if it's the former.

app/src/main/java/com/neeva/app/billing/BillingScreen.kt
```diff
@@ -0,0 +1,106 @@
...
+    val purchases = subscriptionManager.purchasesFlow.collectAsState().value
+    val productDetails = subscriptionManager.productDetailsFlow.collectAsState().value
+    val obfuscatedUserID = subscriptionManager.obfuscatedUserIDFlow.collectAsState().value
+    Timber.tag("KOBE").e("obfuscatedUserID = $obfuscatedUserID")
+    Timber.tag("KOBE").e("purchases = $purchases")
+    Timber.tag("KOBE").e("product details = $productDetails")
+
+    Surface {
+        Box(Modifier.fillMaxSize()) {
+            Column(Modifier.fillMaxSize()) {
```
dan-alcantara-neeva, line 39: Should this be scrollable?

app/src/main/java/com/neeva/app/billing/BillingScreen.kt
```diff
@@ -0,0 +1,106 @@
...
+
+    val purchases = subscriptionManager.purchasesFlow.collectAsState().value
+    val productDetails = subscriptionManager.productDetailsFlow.collectAsState().value
+    val obfuscatedUserID = subscriptionManager.obfuscatedUserIDFlow.collectAsState().value
+    Timber.tag("KOBE").e("obfuscatedUserID = $obfuscatedUserID")
+    Timber.tag("KOBE").e("purchases = $purchases")
+    Timber.tag("KOBE").e("product details = $productDetails")
+
+    Surface {
+        Box(Modifier.fillMaxSize()) {
```
dan-alcantara-neeva, line 38: I think the `Surface` already acts like a `Box` so you can move the modifier to the `Surface` and remove the `Box`.

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
```diff
@@ -0,0 +1,171 @@
...
+                    while (!success && retries <= MAX_RETRIES) {
+                        val delayTime = (1000L * RETRY_WAIT_TIME_MULTIPLIER.pow(retries))
+                            .toLong()
+                        if (billingClientWrapper.retryConnection(delayTime)) {
+                            success = job()
+                        }
+                        retries += 1
+                    }
+                    if (success) {
+                        retries = 0
```
dan-alcantara-neeva, line 146: You never set this back to 0 unless a job succeeds, but you only allow a job to run if `retries <= MAX_RETRIES`.  This means that you'll never try to fire off any of your jobs until the app dies and restarts.  You should figure out an opportunity to reset it without the app dying (like `onResume`).

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
```diff
@@ -0,0 +1,171 @@
...
+            synchronized(activeJobSet) {
+                if (activeJobSet.contains(uniqueJobName)) {
+                    return@launch
+                }
+                activeJobSet.add(uniqueJobName)
+            }
+
+            jobQueueFlow.emit(
+                BillingClientJob(uniqueJobName = uniqueJobName) {
+                    var success = job()
```
dan-alcantara-neeva, line 136: Everything in here about running and retrying the job should be part of the `processJob`  function and not the `queueJob` function.  `BillingClientJob` should just hold a reference to `job`, which would then get used elsewhere.

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -0,0 +1,181 @@
...
+        .map { purchaseList ->
+            purchaseList.any { purchase ->
+                purchase.products.contains(BillingSubscriptionPlanTags.ANNUAL_PREMIUM_PLAN)
+            }
+        }
+
+    val hasMonthlyPremiumPlan: Flow<Boolean> = billingClientController.purchasesFlow
+        .map { purchaseList ->
+            purchaseList.any { purchase ->
+                purchase.products.contains(BillingSubscriptionPlanTags.ANNUAL_PREMIUM_PLAN)
```
dan-alcantara-neeva, line 43: MONTHLY, not ANNUAL.

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -0,0 +1,181 @@
...
+     * @param tag string representing tags associated with offers and base plans.
+     *
+     * @return the eligible offers and base plans in a list.
+     *
+     */
+    private fun retrieveEligibleOffers(
+        offerDetails: MutableList<ProductDetails.SubscriptionOfferDetails>,
+        tag: String
+    ): List<ProductDetails.SubscriptionOfferDetails> {
+        val eligibleOffers = emptyList<ProductDetails.SubscriptionOfferDetails>().toMutableList()
```
dan-alcantara-neeva, line 60: Can you just create a `MutableList` directly?  Creating an `emptyList` and then converting it into a `MutableList` is an unnecessary step

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -0,0 +1,181 @@
...
+            if (offerDetail.offerTags.contains(tag)) {
+                eligibleOffers.add(offerDetail)
+            }
+        }
+
+        return eligibleOffers
+    }
+
+    /**
+     * Calculates the lowest priced offer amongst all eligible offers.
```
dan-alcantara-neeva, line 71: lowest-priced

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -0,0 +1,181 @@
...
+     * @return the offer id token of the lowest priced offer.
+     */
+    private fun leastPricedOfferToken(
+        offerDetails: List<ProductDetails.SubscriptionOfferDetails>
+    ): String? {
+        var offerToken: String? = null
+        var leastPricedOffer: ProductDetails.SubscriptionOfferDetails
+        var lowestPrice = Int.MAX_VALUE
+
+        if (offerDetails.isNotEmpty()) {
```
dan-alcantara-neeva, line 87: The `if` is redundant because the `for` loop won't do anything if `offerDetails` is empty.

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -0,0 +1,181 @@
...
+     *
+     * @param productDetails ProductDetails object returned by the library.
+     * @param purchases List of current [Purchase] objects needed for upgrades or downgrades.
+     * @param billingClient Instance of [BillingClientController].
+     * @param activity [Activity] instance.
+     * @param tag String representing tags associated with offers and base plans.
+     */
+    fun buy(
+        productDetails: ProductDetails,
+        purchases: List<Purchase>?,
```
dan-alcantara-neeva, line 138: existingPurchases?

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -0,0 +1,181 @@
...
+    private fun leastPricedOfferToken(
+        offerDetails: List<ProductDetails.SubscriptionOfferDetails>
+    ): String? {
+        var offerToken: String? = null
+        var leastPricedOffer: ProductDetails.SubscriptionOfferDetails
+        var lowestPrice = Int.MAX_VALUE
+
+        if (offerDetails.isNotEmpty()) {
+            for (offer in offerDetails) {
+                for (price in offer.pricingPhases.pricingPhaseList) {
```
dan-alcantara-neeva, line 89: Can you use `minBy` instead?

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -0,0 +1,181 @@
...
+            productDetails.subscriptionOfferDetails?.let {
+                retrieveEligibleOffers(
+                    offerDetails = it,
+                    tag = tag.lowercase()
+                )
+            }
+        val offerToken = offers?.let { leastPricedOfferToken(it) }
+        val obfuscatedUserID = billingClientController.obfuscatedUserIDFlow.value
+
+        // TODO(kobec): add obfuscatedUserID != null as a condition
```
dan-alcantara-neeva, line 152: Is this a required TODO?  What happens if you haven't accounted for it?

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -0,0 +1,181 @@
...
+        if (purchases.isNullOrEmpty()) {
+            val billingParams = offerToken?.let {
+                billingFlowParamsBuilder(
+                    productDetails = productDetails,
+                    offerToken = it,
+                    obfuscatedUserID = obfuscatedUserID
+                )
+            }
+
+            if (billingParams != null) {
```
dan-alcantara-neeva, line 162: Should this just live inside of the `?.let` block above?  You don't use the `billingParams` anywhere except this call

app/src/main/java/com/neeva/app/billing/billingclient/FetchPurchasesManager.kt
```diff
@@ -0,0 +1,111 @@
...
+        val billingClient = billingClientController.getBillingClient()
+
+        return suspendCoroutine { fetchContinuation ->
+            billingClient?.queryPurchasesAsync(params) { billingResult, purchaseList ->
+                if (
+                    billingResult.responseCode == BillingClient.BillingResponseCode.OK &&
+                    purchaseList.isNotEmpty()
+                ) {
+                    fetchContinuation.resume(true)
+                    // If the BillingClient's last connection is cached, it is possible that this
```
dan-alcantara-neeva, line 42: Move line 41 next to line 47 to keep your comment block together?

app/src/main/java/com/neeva/app/billing/billingclient/FetchPurchasesManager.kt
```diff
@@ -0,0 +1,111 @@
...
+    val purchasesFlow: StateFlow<List<Purchase>> = _purchasesFlow
+
+    suspend fun fetchPurchases(): Boolean {
+        val params = QueryPurchasesParams.newBuilder()
+            .setProductType(BillingClient.ProductType.SUBS)
+            .build()
+        val billingClient = billingClientController.getBillingClient()
+
+        return suspendCoroutine { fetchContinuation ->
+            billingClient?.queryPurchasesAsync(params) { billingResult, purchaseList ->
```
dan-alcantara-neeva, line 36: Can `queryPurchasesAsync` return null?  If not, do a null check on `billingClient` before trying to use it and move your block on line 55 up to here so it's a clear error condition that causes an early exit.

app/src/main/java/com/neeva/app/billing/billingclient/FetchPurchasesManager.kt
```diff
@@ -0,0 +1,111 @@
...
+    override fun onPurchasesUpdated(
+        billingResult: BillingResult,
+        purchases: MutableList<Purchase>?
+    ) {
+        if (
+            billingResult.responseCode == BillingClient.BillingResponseCode.OK &&
+            !purchases.isNullOrEmpty()
+        ) {
+            _purchasesFlow.value = purchases
+            for (purchase in purchases) {
```
dan-alcantara-neeva, line 70: `purchases.forEach` to stay kotlin-y.

app/src/main/java/com/neeva/app/billing/billingclient/FetchPurchasesManager.kt
```diff
@@ -0,0 +1,111 @@
...
+                acknowledgePurchase(purchase)
+            }
+        } else {
+            Timber.e("error code = ${billingResult.responseCode}. ${billingResult.debugMessage}")
+        }
+    }
+
+    /** Perform new subscription purchases' acknowledgement client side. */
+    private fun acknowledgePurchase(purchase: Purchase?) {
+        purchase?.let {
```
dan-alcantara-neeva, line 82: Early exit if `purchase is null`.  Allowing a `null` `purchase` as an argument is weird, though.  Should you do `purchases.filterNotNull()` and make sure this is not null, instead?

app/src/main/java/com/neeva/app/billing/billingclient/FetchPurchasesManager.kt
```diff
@@ -0,0 +1,111 @@
...
+            if (!it.isAcknowledged) {
+                val params = AcknowledgePurchaseParams.newBuilder()
+                    .setPurchaseToken(it.purchaseToken)
+                    .build()
+
+                Timber.e("purchase token acquired: ${it.purchaseToken}")
+
+                billingClientController.getBillingClient()?.acknowledgePurchase(
```
dan-alcantara-neeva, line 90: Early exit if the `billingClient` is null.

app/src/main/java/com/neeva/app/billing/billingclient/FetchProductDetailsManager.kt
```diff
@@ -0,0 +1,58 @@
...
+                .setProductId(SUB_PRODUCT_ID)
+                .setProductType(BillingClient.ProductType.SUBS)
+                .build()
+        )
+        val params = QueryProductDetailsParams.newBuilder()
+            .setProductList(productList)
+        val billingClient = billingClientController.getBillingClient()
+
+        return suspendCoroutine { fetchContinuation ->
+            billingClient
```
dan-alcantara-neeva, line 34: 1. Can `billingClient` change or become null before the `suspendCoroutine` code block fires?  I'm guessing yes if you try to retry a call, so you should move retrieval of `billingClient` to line 34.
2. early exit if `billingClient` is not null.

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientWrapper.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    fun terminateBillingConnection() {
+        // Although this isn't guaranteed be called, if the BillingClient connection is invalid,
+        // a future use of BillingClient (i.e. fetchPurchases or fetchProductDetails) will have an
+        // error. This will trigger the old, invalid billingClient to be destroyed and replaced by a
+        // new one.
+        billingClient?.endConnection()
+    }
+
+    suspend fun retryConnection(delayTime: Long): Boolean {
```
dan-alcantara-neeva, line 64: 1. You can't guarantee that this will be called by only one place, meaning that you could have a new coroutine terminating a `billingClient` used by another coroutine that's already running.  You should mark this as `@NonThreadSafe`
2. Mark relevant functions as `internal` to avoid other callers making unexpected calls.

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientWrapper.kt
```diff
@@ -0,0 +1,82 @@
...
+        delay(delayTime)
```
dan-alcantara-neeva, line 65: Don't pass the `delayTime` into here; use that value where you call `retryConnection`.

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientWrapper.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    private fun createNewBillingClient(): BillingClient {
+        return BillingClient.newBuilder(appContext)
+            .setListener(purchasesUpdatedListener)
+            .enablePendingPurchases()
+            .build()
+    }
+
+    private suspend fun startBillingConnection(): Boolean {
+        return suspendCoroutine { continuation ->
```
dan-alcantara-neeva, line 36: Your `suspendCoroutine` block will never call `continuation.resume` if `billingClient` is null

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
```diff
@@ -0,0 +1,171 @@
...
+        synchronized(activeJobSet) {
+            activeJobSet.remove(billingClientJob.uniqueJobName)
+        }
+    }
+
+    fun launchBillingFlow(activity: Activity, billingParams: BillingFlowParams) {
+        val billingResult = billingClientWrapper.billingClient
+            ?.launchBillingFlow(activity, billingParams)
+
+        if (billingResult == null || billingResult.responseCode != BillingResponseCode.OK) {
```
dan-alcantara-neeva, line 167:
```
if (billingResult?.responseCode != BillingResponseCode.OK)
```

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
```diff
@@ -0,0 +1,171 @@
...
+    private fun queueJobWithRetry(
+        uniqueJobName: String,
+        job: suspend () -> Boolean
+    ) {
+        coroutineScope.launch(dispatchers.io) {
+            synchronized(activeJobSet) {
+                if (activeJobSet.contains(uniqueJobName)) {
+                    return@launch
+                }
+                activeJobSet.add(uniqueJobName)
```
dan-alcantara-neeva, line 131: `add` returns a boolean saying whether or not the item was added.  You can use that and avoid the `contains`.

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
```diff
@@ -0,0 +1,171 @@
...
+
+    // Flows
+    val purchasesFlow: StateFlow<List<Purchase>> = fetchPurchasesManager.purchasesFlow
+    val productDetailsFlow: StateFlow<ProductDetails?> =
+        fetchProductDetailsManager.productDetailsFlow
+    private val _obfuscatedUserIDFlow: MutableStateFlow<String?> = MutableStateFlow(null)
+    val obfuscatedUserIDFlow: StateFlow<String?> = _obfuscatedUserIDFlow
+
+    fun onUserSignedIn() {
+        if (settingsDataModel.getToggleState(SettingsToggle.DEBUG_ENABLE_BILLING).value) {
```
dan-alcantara-neeva, line 72: Early exit if your toggle isn't set correctly.

dan-alcantara-neeva (changes requested):

app/src/main/java/com/neeva/app/billing/BillingScreen.kt
Kokobe: `Box` is actually needed so I can put the close button the TopEnd

app/src/main/java/com/neeva/app/billing/BillingScreen.kt
Kokobe: Yes, but that work I will do in a future PR where I actually implement the Billing Design

app/src/main/java/com/neeva/app/billing/BillingScreen.kt
Kokobe: This is just debugging output. I'll add as a comment to this file that all of this is behind a flag + just a prototype

app/src/main/java/com/neeva/app/billing/BillingScreen.kt
Kokobe: These are all debugging

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
Kokobe: It's required because the back-end needs to map a purchaseToken with a userID. Currently only works on carl's server so thats why I removed the condition for now

app/src/main/java/com/neeva/app/billing/billingclient/FetchPurchasesManager.kt
Kokobe: queryPurchasesAsync does not return anything (`void`), but I'll move the early exit condition higher so its clear

app/src/main/java/com/neeva/app/billing/billingclient/FetchProductDetailsManager.kt
Kokobe: Moved it into the suspendCoroutine!

+ Early exited

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientWrapper.kt
Kokobe: Had to mark the class as `@NotThreadSafe`

app/src/main/java/com/neeva/app/billing/BillingScreen.kt
dan-alcantara-neeva: You already have a BoxScope, don't you?  Look at the definition of Surface

dan-alcantara-neeva: Unless it doesn't pass it down so that you can't set the alignment

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -0,0 +1,163 @@
...
+                purchase.products.contains(BillingSubscriptionPlanTags.ANNUAL_PREMIUM_PLAN)
+            }
+        }
+
+    val hasMonthlyPremiumPlan: Flow<Boolean> = billingClientController.purchasesFlow
+        .map { purchaseList ->
+            purchaseList.any { purchase ->
+                purchase.products.contains(BillingSubscriptionPlanTags.MONTHLY_PREMIUM_PLAN)
+            }
+        }
```
dan-alcantara-neeva, line 45: Add `distinctUntilChanged` here and on line 33?  Avoids the same boolean getting sent out twice

app/src/main/java/com/neeva/app/billing/billingclient/FetchPurchasesManager.kt
```diff
@@ -0,0 +1,128 @@
...
+                return@suspendCoroutine
+            }
+
+            val params = AcknowledgePurchaseParams.newBuilder()
+                .setPurchaseToken(purchase.purchaseToken)
+                .build()
+
+            Timber.e("Purchase token acquired: ${purchase.purchaseToken}")
+
+            billingClientController.getBillingClient()?.acknowledgePurchase(
```
dan-alcantara-neeva, line 108: reuse the `billingClient` from line 95

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
```diff
@@ -0,0 +1,170 @@
...
+        job: suspend () -> Boolean
+    ) {
+        coroutineScope.launch(dispatchers.io) {
+            synchronized(activeJobSet) {
+                val hasBeenAdded = activeJobSet.add(uniqueJobName)
+                if (hasBeenAdded) {
+                    return@launch
+                }
+            }
+            jobQueueFlow.emit(BillingClientJob(uniqueJobName = uniqueJobName, job = job))
```
dan-alcantara-neeva, line 136: Can you just put this in the synchronized block?  You're already early exiting in the opposite case, so you might as well just put the call there.
```
            synchronized(activeJobSet) {
                val hasBeenAdded = activeJobSet.add(uniqueJobName)
                if (!hasBeenAdded) {
                    jobQueueFlow.emit(BillingClientJob(uniqueJobName = uniqueJobName, job = job))
                }
            }
```

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -0,0 +1,163 @@
...
+        offerDetails: List<ProductDetails.SubscriptionOfferDetails>?
+    ): String? {
+        if (offerDetails.isNullOrEmpty()) {
+            Timber.e("OfferDetails is null or empty so could not find an offer token.")
+            return null
+        }
+
+        return offerDetails
+            .associateBy(
+                { it },
```
dan-alcantara-neeva, line 88: Can you just use `it.offerToken`?

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
```diff
@@ -0,0 +1,170 @@
...
+            _obfuscatedUserIDFlow.value = response?.obfuscatedUserID
+        }
+    }
+
+    fun getBillingClient(): BillingClient? {
+        return billingClientWrapper.billingClient
+    }
+
+    fun onCreate() {
+        coroutineScope.launch(dispatchers.io) {
```
dan-alcantara-neeva, line 90: Move this to `init { }` so you have no chance of missing anything from the Flow and so you can use `replay=0, extraBufferCapacity=2`.

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
Kokobe: I cannot because I get this error:
Screen Shot 2022-12-01 at 2 22 58 PM

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
Kokobe: My current MutableSharedFlow defined as:
```
private val jobQueueFlow: MutableSharedFlow<BillingClientJob> = MutableSharedFlow(
        replay = 0,
        extraBufferCapacity = 2
    )
```

But according to https://stackoverflow.com/questions/71309128/mutablesharedflow-difference-between-replay-and-extrabuffercapacity and the docs, it seems that doing replay=0 and extraBufferCapacity=2 is equivalent, no?

Kokobe: I'm going to have to change it to extraBufferCapacity = 3 since I introduced a new job (`acknowledgePurchase`) that can be queued up

dan-alcantara-neeva: It's not equivalent; read the answer (and the comments for the answer) more closely.  `replay` is the number of things that will be replayed to subscribers that started collecting later than the items were added to the queue.  This means that another subscriber will get those `replay` number of items again.  `extraBufferCapacity` is the number of additional items to store that don't get replayed.

Kokobe: Ah but in this case, where there is only 1 subscriber in the `init`, it is the same thing. Thanks for helping me understand that

app/src/main/java/com/neeva/app/billing/BillingScreen.kt
Kokobe: Screen Shot 2022-12-01 at 2 43 15 PM

Yeah unlike Box's parameters that pass down `BoxScope` in the content composable lambda, Surface does not :(

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
```diff
@@ -129,7 +114,7 @@ class BillingClientController(
         coroutineScope.launch(dispatchers.io) {
             synchronized(activeJobSet) {
                 val hasBeenAdded = activeJobSet.add(uniqueJobName)
-                if (hasBeenAdded) {
+                if (!hasBeenAdded) {
```
dan-alcantara-neeva, line 117: Something looks wonky here; I think you were supposed to move the emit line into here instead of just negating the boolean?

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
```diff
@@ -0,0 +1,168 @@
...
+        if (billingResult?.responseCode != BillingResponseCode.OK) {
+            Timber.e("Error in launching billing flow. ${billingResult?.debugMessage}")
+        }
+    }
+
+    fun onUserSignedIn() {
+        if (!settingsDataModel.getToggleState(SettingsToggle.DEBUG_ENABLE_BILLING).value) {
+            return
+        }
+        coroutineScope.launch(dispatchers.io) {
```
dan-alcantara-neeva, line 160: This is called after every attempt to fetch the user, which means that you're allowing this coroutine to fire multiple times in parallel.  Is that intentional, or do you want this job to be done only once?

app/src/test/java/com/neeva/app/billing/BillingClientControllerTest.kt
```diff
@@ -0,0 +1,176 @@
...
+        coroutineScopeRule.scope.advanceUntilIdle()
+
+        billingClientController.queueJobWithRetry("job 1") {
+            testJobMock.job1()
+            true
+        }
+
+        coroutineScopeRule.scope.advanceUntilIdle()
+
+        verify(exactly = 2) { testJobMock.job1() }
```
dan-alcantara-neeva, line 174: Add a `verify(exactly = 1) { ... }` counterpart after line 165

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
Kokobe: I was wrong the first time, but the current logic you see here is right.

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
Kokobe: It is possible for this coroutine to fire multiple apollo calls at once if the `onUserSignedIn()` function is spammed. 

Currently, it is only being called in NeevaUser's fetch() which is only called when the login cookie is valid (Every login or app startup). 

I think that is okay... but if not, I can add state via `SharedPreference`, but managing that can be a separate PR

dan-alcantara-neeva (approved):

PR #1111, opened 2022-11-21
Update Schema and misc graphql

Kokobe:
Preparing: https://github.com/neevaco/neeva-android/pull/1103


dan-alcantara-neeva (approved):

andjd:
@Kokobe The diff here is a mess because the schema.json file switched from two space indents to four.  It would be good if you could fix this to makes the diff more readable.

Kokobe:
@andjd Shoot it auto merged before I could do anything. It looks like the script I ran in the neeva repo generated a file with 4 space indents, I can't control that script and if we modify our json back to 2 indents, we're gonna see this large diff again.

PR #1117, opened 2022-11-28
String update 2022/11/28

dan-alcantara-neeva:
* Update strings files based off of files pulled from Smartling on 2022-11-28.
* Add a script that extracts and puts the files in the right place.


Kokobe (approved):

andjd (changes requested):
The spaces for the sentence separators need to be escaped (`\u0020` for a normal space) as the XML parser will strip leading and trailing whitespace.

dan-alcantara-neeva:
Hrm.  If that were the case then Smartling's export is screwed up...

dan-alcantara-neeva:
It's not even showing up in the Smartling dashboard, for some reason.

dan-alcantara-neeva:
@andjd Gave up on Smartling trying to translate a space character and just said it's not a translatable string in our source file.  PTAL.

andjd:
Question. If the strings are non-translatable, is there any reason to put them in the translated files?

dan-alcantara-neeva:
> Question. If the strings are non-translatable, is there any reason to put them in the translated files?

Not normally, but the way Smartling works is that it takes your original `strings.xml` file (comments, non-translatable strings and all) and then replaces the original text with the translations.  So if I don't do it now, it'll just happen the next time we get translations again.

andjd (approved):

PR #1134, opened 2022-12-07
Better handle the login callback for Custom Tabs

dan-alcantara-neeva:
When the backend finishes the login pathway, it will make the Custom Tab navigate to `neeva://login/cb` with various query parameters that we need parse out to figure out what happened on the backend:
* `sessionKey` = Authentication token
* `finalPath` = Where the user should be navigated to once the login attempt completes
* `retry` = Error code fired when login failed

Changes:
* Add the `LoginCallbackIntentParams` for parsing the `neeva:` URI and its params, which was originally pulled out of `LoginToken.kt`.

* Add corresponding tests for the `LoginCallbackIntentParams`, which were pulled out of and adapted from `LoginTokenTest.kt`.

* Remove the "ios" callback type and always pass in the "android" callback type.  The need for this workaround was fixed 6 months ago by https://github.com/neevaco/neeva/commit/088fba7b3f29cc1c0c36eb0684a0af65e7c1bee2

* Check if a `retry` code is passed back or if a `sessionKey` fails to be sent back.  If either of these happens, inform the user that something went wrong in the most precise way possible based on the available error code.
  * Various error codes can be gleaned from `neeva/auth/challenge/authserver/login_handlers.go` in the backend code.
  * Note that the backend doesn't tell us in the callback whether or not the user was trying to sign up or sign in, meaning that we can't display a more specific error than "Something went wrong".

* Remove the `/?nva` `finalPath` set for Okta logins.  This was apparently added for marketing ages ago and isn't actually used anywhere.

* Fix the login pathway so that it actually waits for the login cookie setting callback to complete before attempting to navigate to a URL.

* Send the user to the URL indicated by the `finalPath` once the login attempt completes.

https://neevaco.slack.com/archives/C02RYG8LY8H/p1670292741375759

### How was this tested?
- [X] I added automated tests

### Manual test cases
* Logged in using a Google account that was previously used to create a Neeva account

https://user-images.githubusercontent.com/96092902/206060333-19884564-0d94-42bb-92e5-2f22927078ce.mp4

* Tried logging in using a Google email address that was previously used to create an account via Okta, saw an error, then logged in via Okta.

https://user-images.githubusercontent.com/96092902/206060577-617f52c7-b0b0-4be7-a9fe-08bcb5846bd8.mp4

## Issues addressed
* Work towards https://github.com/neevaco/neeva-android/issues/791


yozuysal (approved):
Thanks Dan! This all makes sense to me. nva was a puzzle from the beginning. Glad you clarified it :)

PR #1140, opened 2022-12-09
Add Welcome Flow Functionality for Welcome Screen

Kokobe:
Adds functionality for each clickable element in the `WelcomeScreen`. 

* Uses the old Set Default Browser Screen, it needs to be redesigned too (in a future PR)

* Moves `SubscriptionManager` to `WelcomeFlowActivity` and changed its dependencies to only needing `ActivityStarter`
* Adds `queueOnSignIn()` to NeevaUser because the mocks say that once the user signs in via CCT, send the user to different pages based on their `SubscriptionType`.

* Uses the `SubscriptionManager` in `WelcomeFlowActivity` and deletes it from `NeevaActivity`
* The CCT Login functionality is actually not ready yet (cc: @dan-alcantara-neeva ), so I've just built a skeleton around it for when it does work. 

* Figma: https://www.figma.com/file/dylvDPB69JG9U1rzZ3HQ52/Android-Premium?node-id=132%3A5081&t=38oad62CWbI7kqVv-0

### How was this tested?
- [ ] I added automated tests

### Manual test cases
Tapped on everything in the Welcome Screen. See recording

https://user-images.githubusercontent.com/20916043/206813588-2be8a59d-c57b-4abc-9425-2ef48d5a2d6d.mov


app/src/main/java/com/neeva/app/ui/PreviewContainer.kt
```diff
@@ -204,7 +228,8 @@ fun PreviewCompositionLocals(content: @Composable () -> Unit) {
         LocalPopupModel provides previewPopupModel,
         LocalNavHostController provides NavHostController(LocalContext.current),
         LocalSharedPreferencesModel provides previewSharedPreferencesModel,
-        LocalSettingsDataModel provides previewSettingsDataModel
+        LocalSettingsDataModel provides previewSettingsDataModel,
```
dan-alcantara-neeva, line 231: Alphabetize

app/src/main/java/com/neeva/app/userdata/NeevaUser.kt
```diff
@@ -134,6 +148,10 @@ class NeevaUserImpl(
             } else {
                 billingClientController.onUserSignedIn()
                 setUserInfo(userQuery.toUserInfo())
+                jobsToRun.forEach {
+                    it.job()
```
dan-alcantara-neeva, line 152: You should remove each job as you go instead of clearing the whole thing.  That way, if this somehow breaks, you're not going to end up with old tasks getting re-run in the queue.  You can do that by making a copy of the `jobsToRun` and iterating on that, them removing each entry from the original as you go.

app/src/main/java/com/neeva/app/userdata/NeevaUser.kt
```diff
@@ -104,6 +112,12 @@ class NeevaUserImpl(
         return loginToken.isEmpty()
     }
 
+    override fun queueOnSignIn(uniqueJobName: String, job: () -> Unit) {
+        if (jobsToRun.none { it.uniqueJobName == uniqueJobName }) {
+            jobsToRun.add(OnSignedInJob(uniqueJobName = uniqueJobName, job = job))
+        }
+    }
+
```
dan-alcantara-neeva, line 120: Are we guaranteed that on call to `fetch()` will not in parallel with another call to `fetch()`?

app/src/main/java/com/neeva/app/userdata/NeevaUser.kt
```diff
@@ -104,6 +112,12 @@ class NeevaUserImpl(
         return loginToken.isEmpty()
     }
 
+    override fun queueOnSignIn(uniqueJobName: String, job: () -> Unit) {
```
dan-alcantara-neeva, line 115: Is this a fork of the other pattern you were doing before?

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -119,7 +131,16 @@ class WelcomeFlowActivity : AppCompatActivity() {
                         }
                     ) {
                         composable(Destinations.WELCOME.name) {
-                            WelcomeScreen(navigateToSignUp = {})
+                            WelcomeScreen(
+                                navigateToPlans = { Timber.tag("KOBE").e("Navigated to plans") },
```
dan-alcantara-neeva, line 135: Remove log

app/src/main/java/com/neeva/app/welcomeflow/WelcomeScreen.kt
```diff
@@ -146,6 +149,66 @@ internal fun ConsentCheckbox(
     }
 }
 
+/**
+ * UNFINISHED COMPOSABLE. I've only wrote the skeleton, but will be unable to confirm if it works
```
dan-alcantara-neeva, line 153: This is where you actually should add a TODO instead of calling it an unfinished composable.

dan-alcantara-neeva (changes requested):

app/src/main/java/com/neeva/app/userdata/NeevaUser.kt
Kokobe: I could make a `mutex` to ensure this doesn't happen, but this means: 

* If I login with User A
* then immediately sign out 
* and then login with User B,
*  if the fetch() from User A is still running, it will only incorrectly load User A's data.

app/src/main/java/com/neeva/app/userdata/NeevaUser.kt
```diff
@@ -148,10 +148,12 @@ class NeevaUserImpl(
             } else {
                 billingClientController.onUserSignedIn()
                 setUserInfo(userQuery.toUserInfo())
-                jobsToRun.forEach {
-                    it.job()
+
+                val jobsToRunCopy = jobsToRun.toMutableSet()
+                jobsToRunCopy.forEach { onSignedInJob ->
+                    onSignedInJob.job()
```
dan-alcantara-neeva, line 154: nit: Maybe don't call the field `job()` too.

dan-alcantara-neeva (approved):

PR #1147, opened 2022-12-14
[Native login] Ignore the result code and try to parse it anyway

dan-alcantara-neeva:
We can get useful information back in the `result` even though the Activity reports a failure.  We do have to catch a `RuntimeException` thrown by the API to get the data and avoid crashing, though.


dan-alcantara-neeva:
cc @clischeske Going to put this in the repo to help with debugging because it can output more errors than we'd normally get.

Kokobe (approved):

PR #1154, opened 2022-12-16
Replace Custom Tab usage with an internal SingleTabActivity

dan-alcantara-neeva:
Instead of using Chrome Custom Tabs, use our own Activity that simulates showing a WebView in a standalone activity.

* This `SingleTabActivity` uses raw WebLayer calls without using `WebLayerModel` or the other classes we use in the rest of the app because it's meant to be standalone and not trigger any other Neeva-related calls.
  * A lot of this was cribbed from the `NeevaActivity`, `WebLayerModel`, `BaseBrowserWrapper`, and `TabCallbacks`.

* The profile used by the `SingleTabActivity` is unique to the `SingleTabActivity` to avoid conflicting with any existing `Browser` profile.
  * This allows to avoid two different Activities trying to manage the same profile and tabs.
  * This also means that we won't be using a shared Cookie jar, which is fine because this Activity will be used during the login process when the user normally wouldn't have any login cookies set.

*  The `SingleTabActivityViewModel` initializes and maintains the `Browser`.  We have to keep checking if the `Browser` is alive when we use it because a lot of WebLayer calls are made using async callbacks and we have to be make sure the Browser is alive when we try.

* The `SingleTabActivityViewModel` maintains a "tab list" which is basically a stack of Tabs.
  * If a new tab or popup is opened, it is added to the end of the stack.
  * If the user hits back and can't navigate backwards, or if a javascript call causes the active tab to close, the user is sent back to the most recently opened tab in the stack.

* The `SingleTabActivity` uses legacy Android Views to simplify integration with WebLayer (which normally expects Android Views).
  * The UI uses a basic `MaterialToolbar` that displays the name of the page and the current URL.
  * The `ProgressBar` we use has to be hacked a bit to make it render properly without any padding that is built into the style; see the comment for details.

### How was this tested?
- [ ] I added automated tests

### Manual test cases
* Tested the sign in flow with a Google account that has 2FA and was created using Okta (which would cause a failure).
* Tested the sign in flow with a Google account that has no 2FA and was not created using Okta (which succeeds).
* Tested the sign in flow using Okta.

Opening Terms of Service from the Welcome Screen:

https://user-images.githubusercontent.com/96092902/208179020-55f099ca-c15d-40f5-b304-892126cd7a2f.mp4

Using Google account:

https://user-images.githubusercontent.com/96092902/208179475-bedddc97-802b-48c1-bd69-380a80c016c0.mp4

2FA, account was already created using Okta instead of Google login:

https://user-images.githubusercontent.com/96092902/208179628-947b1e9f-b467-4017-93a4-af41a4541c26.mp4

Multiple tab handling (the links open new tabs) and going back and forth:

https://user-images.githubusercontent.com/96092902/208180271-ec0ae6db-1232-475d-8f55-4f81635ee6d4.mp4

Poor 2FA error handling that would happen in CCTs and we should fix:

https://user-images.githubusercontent.com/96092902/208179539-0687a073-6bb4-4e2c-b169-7726038075ef.mp4


app/src/main/AndroidManifest.xml
```diff
@@ -99,8 +99,14 @@
...
-            android:configChanges="orientation|screenSize|screenLayout">
-        </activity>
+            android:configChanges="orientation|screenSize|screenLayout" />
+
+        <activity
+            android:name=".singletabbrowser.SingleTabActivity"
+            android:exported="false"
+            android:theme="@style/Theme.NeevaSearchBrowser.NoActionBar"
+            android:windowSoftInputMode="stateHidden"
+            android:configChanges="orientation|screenSize|screenLayout" />
```
Kokobe, line 109: add: `android:screenOrientation="portrait"`

app/src/main/java/com/neeva/app/singletabbrowser/SingleTabActivity.kt
```diff
@@ -0,0 +1,123 @@
...
+    override fun onBackPressed() {
+        lifecycleScope.launch(Dispatchers.Main) {
+            if (!viewModel.onBackPressed()) {
+                cancelAndFinishActivity()
+            }
+        }
+    }
+
+    private fun cancelAndFinishActivity() {
+        setResult(RESULT_CANCELED)
```
Kokobe, line 112: Handling this result is code to be written later right?

app/src/main/java/com/neeva/app/singletabbrowser/SingleTabActivityViewModel.kt
```diff
@@ -0,0 +1,300 @@
...
+
+    private val isWebLayerLoaded = CompletableDeferred(false)
+    private val tabList = mutableListOf<String>()
+
+    /** IDs of Tabs that have had callbacks registered. */
+    private val registeredTabIds = mutableMapOf<String, TabCallbacks>()
+
+    val titleFlow = MutableStateFlow<String?>(null)
+    val urlFlow = MutableStateFlow<String?>(null)
+    val progressFlow = MutableStateFlow(100)
```
Kokobe, line 57: Should this start at 0?

Kokobe (approved):
lgtm % comments

app/src/main/java/com/neeva/app/singletabbrowser/SingleTabActivity.kt
Kokobe: Where is this callback?

dan-alcantara-neeva: It's part of the Android API: check `Context.startActivityForResult`.  We're not using it right now, though.

app/src/main/java/com/neeva/app/singletabbrowser/SingleTabActivityViewModel.kt
dan-alcantara-neeva: I set it to 100 so that we can avoid having the progress bar show up until it's necessary.

app/src/main/AndroidManifest.xml
dan-alcantara-neeva: We're not limiting this to portrait like we didn't limit how Custom Tabs were opened; we only limit the First Run flow because we don't have designs for that.  If the user is holding their phone in portrait when they open this new Activity, then they'll still be in portrait when it opens.

Issue #1158, opened 2022-12-16
UserInfo returned from backend query keeps returning null for a while


dan-alcantara-neeva:
After logging in, the app performs a `UserInfoQuery` to get the user's info from the backend.  However, it's returning a null response with no exceptions, which is a silent failure on the client.  It's not clear why the backend is returning null, but we should recover better.

![Uploading Screen Shot 2022-12-16 at 10.48.23 AM.png…]()

PR #1159, opened 2022-12-16
Fetch user info when the user opens settings

dan-alcantara-neeva:
Currently, the user's info is only fetched when we detect that the login token has changed, which only happens at startup and when they first log in.  If the fetch fails in either of these scenarios, the user will be unable to properly view their user info in Settings.  Address this by fetching the user's info when they open settings.

### How was this tested?
- [ ] I added automated tests

### Manual test cases
* Added debugging logs, signed in and confirmed that the user's info was updated when Settings was opened.

## Issues addressed
* Partially addresses https://github.com/neevaco/neeva-android/issues/1158, but the backend calls can still return us null incorrectly.


app/src/main/java/com/neeva/app/settings/main/MainSettingsPane.kt
```diff
@@ -18,6 +19,10 @@ import com.neeva.app.ui.PortraitPreviews
 @Composable
 fun MainSettingsPane(settingsController: SettingsController, neevaConstants: NeevaConstants) {
     SettingsPane(settingsController, MainSettingsData(neevaConstants))
+
+    LaunchedEffect(Unit) {
+        settingsController.fetchUserInfo()
```
Kokobe, line 24: Does this protect us against fetch() getting spammed by opening and closing Settings?

dan-alcantara-neeva: Not in the slightest; this just triggers it when the user opens Settings.  A more robust solution would be to have the NeevaUser keep track of when the data was last fetched and only triggering it after a cooldown.  Does that sound useful?

Kokobe (approved):

PR #1161, opened 2022-12-19
Welcome Flow: Implement Login Flow

Kokobe:
Implements all steps of the Login Flow:
* Creates a `LaunchLoginFlow` composable that returns the proper lambda to launch a login flow. 
* When returning to the Neeva app from the intent launched by the login flow, the intent is now handled so that when you return, it returns you to the given screen you specified it to return to. 

* https://www.figma.com/file/dylvDPB69JG9U1rzZ3HQ52/Android-Premium?node-id=0%3A1&t=BnzcsZ7ExyPOdl35-0

* Doc: https://paper.dropbox.com/doc/Welcome-Flow-with-Login-Token--BvMGI7c6hSKKCkQVVfaUrdQ~Ag-eJSSSJXEIVd2uilnQFm1r
* Flowchart: https://www.figma.com/file/SB6HAQfbKjh6kiN7hWW1kc/WelcomeFlow-with-Login-Token?node-id=1%3A667&t=QeZOieojaQu6erJz-0

### How was this tested?
- [ ] I added automated tests

### Manual test cases
See videos

#### Testing with Default Browser set. (Will return to Settings once done)

https://user-images.githubusercontent.com/20916043/208803940-2e197e4b-aa62-45a6-b3a0-c75509acb31a.mov

#### Testing with default browser not set (will send user to ZeroQuery after it is set):

https://user-images.githubusercontent.com/20916043/208803923-98fd387a-2225-4505-ae72-f2ac2596cfaa.mov


dan-alcantara-neeva (changes requested):
Still reviewing, but a first batch of comments to start with.

app/src/main/java/com/neeva/app/NeevaActivity.kt
```diff
@@ -429,18 +440,27 @@ class NeevaActivity : AppCompatActivity(), ActivityCallbacks {
...
-
-                            if (firstRunModel.shouldLogFirstLogin()) {
-                                clientLogger.logCounter(
-                                    LogConfig.Interaction.LOGIN_AFTER_FIRST_RUN, null
-                                )
-                                firstRunModel.setShouldLogFirstLogin(false)
+                            val screenToReturnTo = firstRunModel.getScreenToReturnAfterLogin()
+
+                            if (screenToReturnTo.isEmpty()) {
+                                // Send the user to a URL created by appending the [finalPath] to
```
dan-alcantara-neeva, line 446: I think you're still supposed to send the user to the `finalPath` after they log in.

app/src/main/java/com/neeva/app/NeevaActivity.kt
```diff
@@ -429,18 +440,27 @@ class NeevaActivity : AppCompatActivity(), ActivityCallbacks {
...
+                                // the base Neeva URL.
+                                showBrowser()
+                                uriToLoad = Uri.parse(neevaConstants.appURL).buildUpon()
+                                    .apply { params.finalPath?.let { path(it) } }
+                                    .build()
+
+                                if (firstRunModel.shouldLogFirstLogin()) {
```
dan-alcantara-neeva, line 453: Do we have equivalent logs or should this be pulled out?

app/src/main/java/com/neeva/app/welcomeflow/createaccount/CreateAccountScreen.kt
```diff
@@ -4,55 +4,125 @@
...
+    navigateToSetDefaultBrowser: () -> Unit,
     onShowOtherSignUpOptions: (() -> Unit)? = null,
+    selectedSubscriptionPlanTag: String?,
     onBack: () -> Unit,
 ) {
+    val firstRunModel = LocalFirstRunModel.current
+    val subscriptionManager = LocalSubscriptionManager.current
+    val activity = LocalContext.current as Activity
+    val onPremiumAvailable = {
+        if (
```
dan-alcantara-neeva, line 50: What happens if neither of these is true?  Nothing happens when the button is clicked?

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -72,16 +76,20 @@ class WelcomeFlowActivity : AppCompatActivity() {
     @Inject lateinit var subscriptionManager: SubscriptionManager
 
     private lateinit var setDefaultAndroidBrowserManager: SetDefaultAndroidBrowserManager
+    private lateinit var navHost: NavHostController
 
     private var sendUserToBrowserOnResume: Boolean = false
-    private val sendUserToBrowserOnResumeKey = "sendUserToBrowserOnResume"
+    private val SEND_USER_TO_BROWSER_KEY = "sendUserToBrowserOnResume"
```
dan-alcantara-neeva, line 82: 1. This should be a const in the companion object.
2. Add comments saying these are supposed to be part of the Intent?

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -72,16 +76,20 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
+    private var selectedSubscriptionPlanTag: MutableState<String?> = mutableStateOf(null)
+    private val SUBSCRIPTION_TAG_KEY: String = "selectedSubscriptionPlanTag"
```
dan-alcantara-neeva, line 84: Ditto.

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -210,8 +234,23 @@ class WelcomeFlowActivity : AppCompatActivity() {
         billingClientController.onDestroy()
     }
 
+    override fun onNewIntent(intent: Intent?) {
+        super.onNewIntent(intent)
+        val screenToReturnTo = firstRunModel.getScreenToReturnAfterLogin()
+        if (screenToReturnTo.isEmpty() || screenToReturnTo == Destinations.WELCOME.name) {
```
dan-alcantara-neeva, line 240: Make this thing a `when(screenToReturnTo)` .

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -59,6 +62,7 @@ class WelcomeFlowActivity : AppCompatActivity() {
             CREATE_ACCOUNT_WITH_GOOGLE,
             CREATE_ACCOUNT_WITH_OTHER
         }
+        val FinishWelcomeFlowOnLoginReturn = "FINISHED_WELCOME_FLOW"
```
dan-alcantara-neeva, line 65: const val, all caps underscored

dan-alcantara-neeva (commented):
Pass 2 while I'm looking at strings

app/src/main/java/com/neeva/app/welcomeflow/LaunchLoginFlow.kt
```diff
@@ -0,0 +1,73 @@
...
+    val params = LaunchLoginFlowParams(
+        provider = provider ?: NeevaUser.SSOProvider.GOOGLE,
+        signup = false
+    )
+
+    val resultLauncher = rememberLauncherForActivityResult(
+        ActivityResultContracts.StartActivityForResult()
+    ) { result: ActivityResult ->
+        firstRunModel.handleLoginActivityResult(context, result, params) {
+            neevaUser.queueOnSignIn(uniqueJobName = "Welcome Flow: onSuccessfulSignIn") {
```
dan-alcantara-neeva, line 45: Should this logic live in a `Model` somewhere instead of a UI `@Composable`?

app/src/main/java/com/neeva/app/browsing/RegularBrowserWrapper.kt
```diff
@@ -159,8 +160,11 @@ class RegularBrowserWrapper(
         WebLayer.createBrowserFragment(NON_INCOGNITO_PROFILE_NAME, PERSISTENCE_ID)
 
     override fun registerBrowserCallbacks(browser: Browser): Boolean {
+        Timber.tag("KOBE").e("registerBrowserCallbacks()")
+        // TODO(kobec): use part of the initializeCookieManager() here to set cookie in Browser from SharedPref value
```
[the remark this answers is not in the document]
Kokobe: Ah this is an old TODO, this already happens because in MainActivity, the loginToken value is updated/set and the browser cookie jar is updated/set (see `handleLoginIntentIfExists()`).

app/src/main/java/com/neeva/app/MainActivity.kt
```diff
@@ -24,29 +32,82 @@ import timber.log.Timber
...
+
+        if (activityClass == null || screenToReturnTo.isEmpty()) {
+            return false
+        }
+
+        loginToken.updateCachedCookie(sessionKey)
+        loginToken.updateBrowserCookieJarWithCachedCookie()
+        firstRunModel.queueFetchNeevaInfo()
+
+        val newIntent = Intent(this@MainActivity, activityClass).setAction(Intent.ACTION_MAIN)
```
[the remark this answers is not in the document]
Kokobe: AFAIK the login intent that opens our app originates from the Neeva login website, which would require the Back-end to be modify the URI string to store new params in.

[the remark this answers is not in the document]
Kokobe: The problem I'm trying to solve is how to return to the client back to a certain `Activity` and `Screen` after Login. The specific `Activity` and `Screen` is decided by the client and where the login was triggered from. 

For example: 
- login from the WelcomeFlow in FirstRun should return you to the WelcomeFlowActivity PLANS screen.

As I wrote this I realize that all logins reroute to the WelcomeFlowActivity and the `screen` is decided by the logged in UserInfo.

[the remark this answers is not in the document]
Kokobe: It doesn't sound necessary to me to pass the `client data` (activity + screen to return to after login) to the Neeva website and then have the website send an Intent opening our app containing the right screen and activity to open... because:
* it requires back-end rewiring
* if I keep this Activity/Screen logic on the client, I can ensure that the Activity/Screen I want to return to definitely exists.
* The Activity/Screen to open is dependent on the information only found on the Client-side (Default Browser, Plans selection, if the user can purchase a subscription on this Android device, if the user already has premium)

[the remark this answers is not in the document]
Kokobe: I'll go the hard-coded enum route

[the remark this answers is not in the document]
Kokobe: After looking into the hard-coded enum route, I don't think it is needed. 

In` MainActivity.kt`, I use `handleLoginIntentIfExists()` to return false when something is off and I want to just normally launch `NeevaActivity`:
```
// in handleLoginIntentIfExists():
val activityClass = getActivityClassToOpen(activityNameToReturnTo)

if (activityClass == null || screenToReturnTo.isEmpty()) {
    return false
}
```
I return null in getActivityClassToOpen() if the sharedPreference value doesn't match the names of WelcomeFlowActivity or NeevaActivity (aka. if we renamed either class to a name that doesn't match the sharedPref, this function would return null)

This snippet will return false if the SharedPref values are outdated which delegates our launching logic to opening the regular NeevaActivity.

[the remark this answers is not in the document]
dan-alcantara-neeva: Yeah, wasn't suggesting passing the client navigation state to the backend.  Was thinking that for other use cases, we'd want to use Intents to pass data around instead of using these SharedPreferences.  If we didn't have to go out to the single-tabbed Activity in some cases or Intent out to the native google login pathway, I imagine we could have avoided most of this...

app/src/main/java/com/neeva/app/NeevaActivity.kt
```diff
@@ -429,18 +440,27 @@ class NeevaActivity : AppCompatActivity(), ActivityCallbacks {
...
-                            uriToLoad = Uri.parse(neevaConstants.appURL).buildUpon()
-                                .apply { params.finalPath?.let { path(it) } }
-                                .build()
-
-                            if (firstRunModel.shouldLogFirstLogin()) {
-                                clientLogger.logCounter(
-                                    LogConfig.Interaction.LOGIN_AFTER_FIRST_RUN, null
-                                )
-                                firstRunModel.setShouldLogFirstLogin(false)
+                            val screenToReturnTo = firstRunModel.getScreenToReturnAfterLogin()
```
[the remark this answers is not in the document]
Kokobe: I've added a function in MainActivity to confirm that the a given screen matches a corresponding Activity `getScreenToOpen()`

app/src/main/java/com/neeva/app/NeevaActivity.kt
Kokobe: Yeah you're right this should be pulled out

app/src/main/java/com/neeva/app/welcomeflow/createaccount/CreateAccountScreen.kt
Kokobe: Great callout. It should navigate to the default browser -- fixed this!

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
Kokobe: I've changed this to be one of the Destinations Enum to be more consistent with the screen names. It doesn't actually have a composable screen linked to it, it's just used as a constant string

app/src/main/java/com/neeva/app/welcomeflow/LaunchLoginFlow.kt
Kokobe: The core benefit here would be organization right? 

I can do that

app/src/main/java/com/neeva/app/NeevaActivity.kt
```diff
@@ -429,18 +440,27 @@ class NeevaActivity : AppCompatActivity(), ActivityCallbacks {
...
+                                    clientLogger.logCounter(
+                                        LogConfig.Interaction.LOGIN_AFTER_FIRST_RUN, null
+                                    )
+                                    firstRunModel.setShouldLogFirstLogin(false)
+                                }
+                            } else {
+                                // There are no other cases where you want to show a screen other
+                                // than Settings when processing a Login Intent in NeevaActivity.
+                                showSettings()
+                                firstRunModel.clearDestinationsToReturnAfterLogin()
```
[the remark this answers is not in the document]
Kokobe: If the SharedPreference values are outdated, I clear it in MainActivity.kt

If they are done with the SharedPref values, in `onNewIntent()` in NeevaActivity.kt and WelcomeFlowActivity.kt , this function will be called. 

I think those are all the cases where I should be clearing them. You'll see the update in my next commit!

app/src/main/java/com/neeva/app/welcomeflow/LaunchLoginFlow.kt
Kokobe: Oh so because of `rememberLauncherForActivityResult`, I can't call this in a normal function in a Model (that function would have to have a `@Composable` tag on top of it anyways. 

I think keeping it the current way is actually cleaner though.

Kokobe: Nevermind, going to move this logic to FirstRunModel

app/src/main/java/com/neeva/app/welcomeflow/createaccount/CreateAccountScreen.kt
Kokobe: After testing on release, I realize that it's okay if neither of those conditions are true. The user will already be routed to either Settings or the SetDefaultBrowser screen. 

The Google Play Subscription Bottom sheet should only pop up if the user selected a non-free Subscription option.

app/src/main/java/com/neeva/app/welcomeflow/LaunchLoginFlow.kt
dan-alcantara-neeva: Composables actually have a `rememberCoroutineScope` function that you can use.  It and its jobs would, however, get canceled as soon as the Composable leaves the screen.

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -129,12 +129,14 @@ class SubscriptionManager(
...
-    fun buy(
-        productDetails: ProductDetails,
-        existingPurchases: List<Purchase>?,
-        activity: Activity,
-        tag: String
-    ) {
+    fun buy(activity: Activity, tag: String) {
+        val productDetails = productDetailsFlow.value
+        if (productDetails == null) {
+            Timber.e("Unable to launch Purchase Flow because product details is null.")
```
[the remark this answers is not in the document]
dan-alcantara-neeva: Missed a comment @Kokobe

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
```diff
@@ -341,6 +367,81 @@ class FirstRunModel internal constructor(
...
+            )
+        SharedPrefFolder.FirstRun
+            .ScreenToReturnToAfterLogin.set(
+                sharedPreferencesModel,
+                activityReturnParams.screenToReturnTo
+            )
+
+        val offers = subscriptionManager.productDetailsFlow.value?.subscriptionOfferDetails
+
+        neevaUser.queueOnSignIn(uniqueJobName = "Welcome Flow: onSuccessfulSignIn") {
```
dan-alcantara-neeva, line 395: Does this hard-coded string const get used anywhere else?

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
```diff
@@ -341,6 +367,81 @@ class FirstRunModel internal constructor(
...
+            // TODO(kobec): Add subscription source check to ensure users who subscribed to a
+            //  Non-Google-Play source cannot purchase a subscription.
+
+            // TODO(kobec): Check purchases too!
+            if (
+                subscriptionType != null &&
+                subscriptionType != SubscriptionType.Basic &&
+                !offers.isNullOrEmpty()
+            ) {
+                coroutineScope.launch(dispatchers.main) {
```
dan-alcantara-neeva, line 407: Does it make sense to pull this `coroutineScope.launch` out?

app/src/main/java/com/neeva/app/userdata/NeevaUser.kt
```diff
@@ -121,9 +125,8 @@ class NeevaUserImpl(
     }
 
     override fun queueOnSignIn(uniqueJobName: String, job: () -> Unit) {
-        if (jobsToRun.none { it.uniqueJobName == uniqueJobName }) {
-            jobsToRun.add(OnSignedInJob(uniqueJobName = uniqueJobName, job = job))
-        }
+        jobsToRun.removeIf { it.uniqueJobName == uniqueJobName }
```
dan-alcantara-neeva, line 128: This is a change of behavior, right?  This means that the job will always get removed and put to the back of the queue.  Is that OK?

app/src/main/java/com/neeva/app/welcomeflow/createaccount/CreateAccountScreen.kt
```diff
@@ -4,55 +4,125 @@
...
+            selectedSubscriptionPlanTag != null &&
+            selectedSubscriptionPlanTag != BillingSubscriptionPlanTags.FREE_PLAN
+        ) {
+            subscriptionManager.buy(activity, selectedSubscriptionPlanTag)
+        }
+    }
+
+    val onPremiumUnavailable = { }
+
+    val (activityToReturnTo, screenToReturnTo) =
```
dan-alcantara-neeva, line 58: You should be using the ActivityReturnParams here instead of continuing to use your Pair

app/src/main/java/com/neeva/app/welcomeflow/createaccount/CreateAccountScreen.kt
```diff
@@ -4,55 +4,125 @@
...
+        if (setDefaultAndroidBrowserManager.isNeevaTheDefaultBrowser()) {
```
dan-alcantara-neeva, line 59: Should this logic and the previous lambda be moved, like into `FirstRunModel`?

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -210,8 +232,19 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
     }
 
+    override fun onNewIntent(intent: Intent?) {
+        super.onNewIntent(intent)
+        when (val screenToReturnTo = firstRunModel.getScreenToReturnToAfterLogin()) {
+            "", Destinations.WELCOME.name -> { }
+            Destinations.FINISH_WELCOME_FLOW.name -> { sendUserToBrowser() }
+            else -> { navHost.navigate(screenToReturnTo) }
+        }
+        firstRunModel.clearDestinationsToReturnAfterLogin()
```
dan-alcantara-neeva, line 242: You should do this immediately after `getScreenToReturnToAfterLogin` so that if the app crashes in between the two lines, the user won't get stuck in a crashloop when the SharedPref is read again.

app/src/main/java/com/neeva/app/welcomeflow/LaunchLoginFlow.kt
```diff
@@ -0,0 +1,49 @@
...
+    val context = LocalContext.current
+
+    val params = LaunchLoginFlowParams(
+        provider = provider ?: NeevaUser.SSOProvider.GOOGLE,
+        signup = false
+    )
+
+    val resultLauncher = rememberLauncherForActivityResult(
+        ActivityResultContracts.StartActivityForResult()
+    ) { }
```
dan-alcantara-neeva, line 34: Is there nothing to do when the `ActivityResult` comes back?

app/src/main/java/com/neeva/app/welcomeflow/LaunchLoginFlow.kt
```diff
@@ -0,0 +1,49 @@
...
+import androidx.activity.result.contract.ActivityResultContracts
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.platform.LocalContext
+import com.neeva.app.LocalFirstRunModel
+import com.neeva.app.firstrun.ActivityReturnParams
+import com.neeva.app.firstrun.LaunchLoginFlowParams
+import com.neeva.app.userdata.NeevaUser
+
+@Composable
+fun LaunchLoginFlow(
```
dan-alcantara-neeva, line 17: This is a Composable that returns something and doesn't draw anything; I think Android Studio warnings would tell you that it should be lowercased.  Can you check?

app/src/main/java/com/neeva/app/userdata/SessionToken.kt
```diff
@@ -92,10 +92,7 @@ abstract class SessionToken(
 
     private var requestJob: Job? = null
 
-    fun initializeCookieManager(browser: Browser, requestCookieIfEmpty: Boolean = false) {
-        val cookieManager = browser.takeIfAlive()?.profile?.cookieManager ?: return
-        weakBrowser = WeakReference(browser)
-
+    fun updateBrowserCookieJarWithCachedCookie() {
         coroutineScope.launch(dispatchers.main) {
             // https://github.com/neevaco/neeva-android/issues/1043
```
dan-alcantara-neeva, line 97: This entire comment should go to the callsite and not here.  You're intentionally calling this function somewhere else and the comment doesn't make sense in that situation.

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
Kokobe: Nope, it's just a job name (not UI to be translated)

dan-alcantara-neeva: Not the reason I was asking: if you had to use this string somewhere else to (say) check if a job was enqueued, then I would have asked for it to get put into a `const val` somewhere.

app/src/main/java/com/neeva/app/welcomeflow/createaccount/CreateAccountScreen.kt
Kokobe: I feel like having the navigation logic in the UI is actually cleaner than storing it in the model. It makes it more intuitive and straight forward what the navigation behavior is going to be at this particular screen

I also think this is no different than passing a navigateToSettings() lambda and using it when a button is pressed or an screen state has occurred. WDYT?

Kokobe: Otherwise all of this logic (which is unique per screen) is going to be stuck into a Model and (not going to get reused as well).

app/src/main/java/com/neeva/app/welcomeflow/LaunchLoginFlow.kt
Kokobe: Originally I had a `neevaUser.queueOnSignIn()` queue a job, but I realized that creates a race condition in which:
the ActivityResult comes back but the `NeevaUser` data has already been fetched and signed in (so the job queued will never happen). 


Since I already have `neevaUser.queueOnSignIn()`, I don't really have anything to do when ActivityResult comes back

app/src/main/java/com/neeva/app/welcomeflow/LaunchLoginFlow.kt
Kokobe: Ah you were right, I changed it

app/src/main/java/com/neeva/app/welcomeflow/LaunchLoginFlow.kt
Kokobe: Ah yeah so the rememberCoroutineScope wouldn't be useful since the screen might change due to the app getting refreshed because of OS memory prioritization

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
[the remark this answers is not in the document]
Kokobe: Ideally the `buy` function wouldn't be called if the product details == null. 

The Purchase Subscription screen which will trigger the buy function will only be showed if the product details != null. I think that's enough gating to make sure this doesn't happen? My Timber log here was mainly just defensive coding for the developer in case they were trying to use this without checking `productDetails`

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
Kokobe: Yeah

app/src/main/java/com/neeva/app/userdata/NeevaUser.kt
Kokobe: Yeah that's fine. I don't want an old job to be running in the queue because:

If the user launches the flow and then decides not to login and then they decide to login again but with different parameters (like a different subscription plan choice), the old job would stay in the queue and the new job with the right parameters would be ignored. 

HOWEVER, I do realize now that this job would still get executed if they decided to not login in until later... which means navigation to the WelcomeFlowActivity can happen while in NeevaActivity which will crash the app

app/src/main/java/com/neeva/app/welcomeflow/createaccount/CreateAccountScreen.kt
Kokobe: Going to put the logic in FirstRunModel in a follow-up PR!

dan-alcantara-neeva (approved):

PR #1174, opened 2023-01-03
String drop 2023-01-03

dan-alcantara-neeva:
Update strings from the last week of December.


app/src/main/res/values-de-rDE/strings.xml
```diff
@@ -988,7 +989,7 @@
     <string name="welcomeflow_annual_premium_plan">Premium\njährlich</string>
 
     <!-- [WelcomeFlow] Name of the Monthly Premium Subscription Plan. -->
-    <string name="welcomeflow_monthly_premium_plan">Monatlich Jährlich</string>
+    <string name="welcomeflow_monthly_premium_plan">Premium\nmonatlich</string>
```
andjd, line 992: do we want hardcoded newlines in here?

app/src/main/res/values-en-rGB/strings.xml
```diff
@@ -988,7 +989,7 @@
     <string name="welcomeflow_annual_premium_plan">Premium\nAnnual</string>
 
     <!-- [WelcomeFlow] Name of the Monthly Premium Subscription Plan. -->
-    <string name="welcomeflow_monthly_premium_plan">Monthly Annual</string>
+    <string name="welcomeflow_monthly_premium_plan">Premium\nMonthly</string>
```
andjd, line 992: Same

andjd (approved):

app/src/main/res/values-de-rDE/strings.xml
dan-alcantara-neeva: Removed from Smartling's memory, as per conversation on Slack:
https://neevaco.slack.com/archives/C02RYG8LY8H/p1672770627823909

tl;dr: iOS hard-coded the newlines in and Smartling gives them back to us even though we don't add a `\n` ourselves.  Removed the newline in our translations, pending input from Hwi.

app/src/main/res/values-en-rGB/strings.xml
dan-alcantara-neeva: Same as above

PR #1178, opened 2023-01-05
Adjust build script for Bee

dan-alcantara-neeva:
Use a shared script in the main neeva repo for building.

Depends on https://github.com/neevaco/neeva/pull/85854, which makes a shared build script that we'll be using for Bee and the Browser.


chung-lai (approved):
LGTM after checking out bee script and the neeva repo script change 🙏

PR #1184, opened 2023-01-08
Use WelcomeFlow for all Login paths

Kokobe:
Uses `WelcomeFlowActivity` for login from Settings and `SpacesIntro` and FirstRun. Release users will not see the Plans screen in this flow because billing is disabled by default.

More improvements: 
* Add signInOnly support to WelcomeFlowActivity (depending on whether `signInOnly` is true or false, it starts WelcomeFlowActivity on a different screen)
* Use WelcomeFlow instead of FirstRunActivity for login in the app
* Add support in PlansScreen to omit Free Plan
* Polish sign in from Spaces
* Remove navigation transitions for sign-in flow
* Get onClick for ProfileRow to refresh whenever neeva user info changes
* Make email promo checkbox actually opt users out of emails
* Remove `SingleTabActivity` from Android Recents when finished with it
* Exclude `WelcomeFlowActivity` from recents
* `IMPORTANT` changes which users will see the `WelcomeFlowActivity` based on `FirstRunDone` SharedPreference: see `MainActivity.kt` for details
* Loading the `finalPath` in our Browser after a successful login.

* https://www.figma.com/file/dylvDPB69JG9U1rzZ3HQ52/Android-Premium?node-id=0%3A1&t=1NHhtmFIyibxIQlO-0
* Discussion on SpacesIntro sign-in/sign-up behavior: https://neevaco.slack.com/archives/C03PQUPJJDP/p1672957695253199

### How was this tested?
- [x] I added automated tests

### Manual test cases
Also manually testing the following test cases in this doc: https://paper.dropbox.com/doc/WelcomeFlowActivity-Manual-Test-Cases--BwhdlrV0WQUQ7GgRuAZAV3uvAg-jcV6Hmx8l6E2PVigflRk3

Sign-in from FirstRun -> shows set default browser only (when Billing disabled)
Sign-in from Settings -> Shows Plans (when Billing enabled) + Clicking on ProfileRow after sign-in shows ProfilePane
Sign-in from SpacesIntro -> does not show Plans (when Billing enabled)
Sign-up from SpacesIntro -> show Plans (when Billing enabled)

Sign-in from FirstRun -> shows set default browser only (when Billing disabled)

https://user-images.githubusercontent.com/20916043/211223794-828f2a66-29da-4ac0-9b96-e48b359d9c15.mp4

Sign-in from Settings -> Shows Plans (when Billing enabled) + Clicking on ProfileRow after sign-in shows ProfilePane

https://user-images.githubusercontent.com/20916043/211250235-00aa1408-b290-47fa-993b-7f7edbb4791a.mp4

Sign-in from SpacesIntro -> does not show Plans (when Billing enabled) + Sign in dismisses bottom sheet

https://user-images.githubusercontent.com/20916043/211249275-d4c15471-e909-45d2-88b2-655ce30e7e57.mp4

Sign-up from SpacesIntro -> show Plans (when Billing enabled) + Sign in dismisses bottom sheet

https://user-images.githubusercontent.com/20916043/211249284-5cf4c904-afe2-486b-a50f-5a17357d7994.mp4


app/src/androidTest/java/com/neeva/app/firstrun/WebSignInFlowTest.kt
```diff
@@ -4,216 +4,210 @@
...
     }
 
     @Test
     fun signInUsingMicrosoft() {
         androidComposeRule.apply {
+            waitFor { multiActivityTestRule.getNeevaActivity() != null }
+            waitFor { activity is NeevaActivity }
+        }
+
+        (androidComposeRule as AndroidComposeTestRule<TestRule, NeevaActivity>).apply {
```
dan-alcantara-neeva, line 114: lol well that's one way to work around that.  For future reference, when you're doing one of these casts in production app code, you should be using `as?` to see if the cast is actually valid.  Otherwise it's just like doing an `!!` on a nullable thing and hoping it's correct.

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -288,6 +325,63 @@ class WelcomeFlowActivity : AppCompatActivity() {
                 screenToReturnTo = screenToReturnToAfterWelcomeFlow
             )
         }
+
+        val signInOnly = intent.extras?.getBoolean(SIGN_IN_ONLY)
+        val billingEnabled = isBillingEnabled()
+        startDestination = when {
```
dan-alcantara-neeva, line 331: Add a comment to describe how each of these destinations make sense given the conditional.  I'm not following why `billingEnabled` is the condition that sends you to the plans or the account creation screen...  Would also help to clarify what `signInOnly` means.

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -288,6 +325,63 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
+                ) {
+                    navHost.navigate(startDestination) {
+                        launchSingleTop = true
+
+                        popUpTo(startDestination)
+                    }
+                }
+            }
+
+            // No valid screens to return to, so do nothing.
```
dan-alcantara-neeva, line 378: Can you clarify what "doing nothing" means here?

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -308,18 +402,7 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
-        when {
-            Destinations.values().any { it.name == screenToReturnToAfterLogin } -> {
-                navHost.navigate(screenToReturnToAfterLogin)
-            }
-
-            screenToReturnToAfterLogin == FINISH_WELCOME_FLOW -> { finishWelcomeFlow() }
-
-            else -> { }
-        }
+        setupFirstScreen()
```
dan-alcantara-neeva, line 405: Calling `setupFirstScreen` after `onNewIntent` is weird.  `onNewIntent` is called after the app has already been opened, so it's not really the first screen anymore.

app/src/main/java/com/neeva/app/welcomeflow/PlansScreen.kt
```diff
@@ -113,6 +137,9 @@ fun PlansScreenContent(
     modifier: Modifier
 ) {
     var selectedTabIndex by remember { mutableStateOf(initialTabIndex) }
+    if (selectedTabIndex > subscriptionPlans.size || selectedTabIndex < 0) {
```
dan-alcantara-neeva, line 140: Use `.coerceIn` to force it in bounds, instead.

dan-alcantara-neeva (changes requested):

app/src/main/java/com/neeva/app/MainActivity.kt
```diff
@@ -99,19 +102,25 @@ class MainActivity : AppCompatActivity() {
...
             return false
         }
 
         loginToken.updateCachedCookie(params.sessionKey)
         // Since it is possible that we don't start the NeevaActivity, we need to make sure that the
         // browser cookie jar has the new cached cookie value.
-        loginToken.updateBrowserCookieJarWithCachedCookie()
+        loginToken.updateBrowserCookieJarWithCachedCookie(
+            onFailure = {
+                // Tell the user something went wrong and that they couldn't be
```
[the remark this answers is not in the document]
Kokobe: The only exception where I would load the neeva home page is if a sign-in failed or you sign in from the browser (both of which I'll double check that it will work)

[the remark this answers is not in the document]
dan-alcantara-neeva: That's what the API says: load the final path after you finish login.  Final path is almost always the home page.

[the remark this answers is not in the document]
dan-alcantara-neeva: In the case where the user started trying to log in from a search result page, then the final path _should_ be the search result page, but the app wasn't set up that way.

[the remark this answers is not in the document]
Kokobe: Offline conversation: decided that changing the behavior of how `finalPath` interacts with what we open up in the tab was:

1. Out of scope of Premium Subscriptions
2. Can muck with the API and how the back-end operates
3. Handling the 10% of users who don't know that the website they want to save to Spaces is in a different tab is an edge case

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
Kokobe: Will rehash in the code comments too:

`billingEnabled` is a boolean variable that is dependent on whether the user is able to purchase Premium or not. If they can't purchase Premium then they shouldn't be shown the `Plans` screen. (also reflected in the figma). I'll change the name of this variable to `isPremiumPurchaseAvailable`

the signInOnly variable determines if a user wants to use this activity only for signing in.

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
Kokobe: Updated the comment to:
There are no valid screens to return to, so no manual setting of which screen to start on is needed.

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
Kokobe: I'm basically just trying to change the starting screen of this activity. I'll rename it to that

app/src/main/java/com/neeva/app/NeevaActivity.kt
```diff
@@ -414,33 +414,18 @@ class NeevaActivity : AppCompatActivity(), ActivityCallbacks {
             ACTION_SHOW_SCREEN -> {
                 switchToRegularProfile()
                 processScreenToNavigateTo()
+                uriToLoad = intent.data
+                // Don't automatically show the Browser
```
dan-alcantara-neeva, line 418: // Set onLoadStarted to an empty lambda so that the tab is opened by the user is sent to the correct screen.

app/src/main/java/com/neeva/app/NeevaActivity.kt
```diff
@@ -414,33 +414,18 @@ class NeevaActivity : AppCompatActivity(), ActivityCallbacks {
...
+                onLoadStarted = { }
             }
 
             Intent.ACTION_VIEW -> {
                 switchToRegularProfile()
-
+                // Logging if we are the default browser when a user views a URL with this app.
```
dan-alcantara-neeva, line 424: // Logs if the app is set as the system's default browser when we are asked to handle a VIEW Intent.

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -327,16 +315,26 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
-        val billingEnabled = isBillingEnabled()
+        val canPurchasePremium = isPremiumPurchaseAvailable()
         startDestination = when {
             firstRunModel.mustShowFirstRun() -> Destinations.WELCOME.name
-            billingEnabled && signInOnly == false -> Destinations.PLANS.name
-            !billingEnabled && signInOnly == false -> Destinations.CREATE_ACCOUNT_WITH_GOOGLE.name
+
+            // If a user wants to sign up (which is different from sign-in), show Premium plans
+            // before sign-up.
+            canPurchasePremium && signInOnly == false -> Destinations.PLANS.name
```
dan-alcantara-neeva, line 324: How does `sign up` correspond to your `signInOnly` field?  I still think it's named in a confusing way

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -327,16 +315,26 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
+
+            // If a user cannot purchase Premium, skip to signup
+            !canPurchasePremium && signInOnly == false ->
+                Destinations
```
dan-alcantara-neeva, line 328: 1. Formatting is weird.
2. If you have to break a new line like this, add braces

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -179,29 +179,23 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
-                                navigateToSignIn = {
-                                    navHost.navigate(Destinations.SIGN_IN.name)
-                                },
-                                navigateToSetDefaultBrowser = {
-                                    navHost.navigate(Destinations.SET_DEFAULT_BROWSER.name)
-                                }
+                                isPremiumPurchaseAvailable = isPremiumPurchaseAvailable(),
+                                navigateToPlans = welcomeFlowNavModel::showPlans,
+                                navigateToSignIn = welcomeFlowNavModel::showSignIn,
+                                navigateToSetDefaultBrowser =
```
dan-alcantara-neeva, line 185: Indentation is weird; just use braces and call the function normally

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -179,29 +179,23 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
-                                loginFlowNavModel::navigateToCreateAccountWithGoogle,
-                                navigateToSignIn =
-                                loginFlowNavModel::navigateToSignIn,
+                                onContinue = getOnContinueInPlansScreen(),
+                                navigateToSignIn = welcomeFlowNavModel::showSignIn,
                                 saveSubscriptionPlanChoice = { tag ->
                                     selectedSubscriptionPlanTag = tag
                                 },
                                 onBack = onBack,
+                                isSignedOut = !neevaUser.isSignedOut(),
```
dan-alcantara-neeva, line 198: This conditional looks backwards?  Why are you negating it?

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -238,18 +232,16 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
                                 onPremiumAvailable = {
                                     subscriptionManager.buy(
                                         this@WelcomeFlowActivity,
                                         selectedSubscriptionPlanTag
                                     )
                                 },
                                 onShowOtherSignUpOptions =
-                                loginFlowNavModel::navigateToCreateAccountWithOther,
-                                navigateToSignIn = { navHost.navigate(Destinations.SIGN_IN.name) },
+                                welcomeFlowNavModel::showCreateAccountWithOther,
```
dan-alcantara-neeva, line 243: Use braces instead of `::` notation

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
Kokobe: Agh that was a typo, thanks for catching that

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
Kokobe: Ah you can either start the WelcomeFlowActivity as signing up or signing-in. Mutually exclusive XOR, but I see how it is confusing. I'll make separate variables based on signInOnly:

```
val usingActivityForSignUp = !signInOnly == false
val usingActivityForSignIn = signInOnly
```

dan-alcantara-neeva: Don't those two fields evaluate to the same value?

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -314,19 +316,20 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
+        val usingActivityForSignIn = intent.extras?.getBoolean(SIGN_IN_ONLY) ?: false
+        val usingActivityForSignUp = !usingActivityForSignIn
         val canPurchasePremium = isPremiumPurchaseAvailable()
         startDestination = when {
             firstRunModel.mustShowFirstRun() -> Destinations.WELCOME.name
 
             // If a user wants to sign up (which is different from sign-in), show Premium plans
             // before sign-up.
-            canPurchasePremium && signInOnly == false -> Destinations.PLANS.name
+            canPurchasePremium && usingActivityForSignUp -> Destinations.PLANS.name
```
dan-alcantara-neeva, line 327: Combine these two under the `usingActivityForSignup` check:

```
usingActivityForSignUp -> {
    if (canPurchasePremium) {
        Destinations.PLANS.name
    } else {
        Destinations.CREATE_ACCOUNT_WITH_GOOGLE.name
    }
}
```

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
Kokobe: SignUp means:
- Show Plans
- Show Sign Up screen
- Show set default browser if they chose a non-free Plan

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
```diff
@@ -114,14 +115,8 @@ class FirstRunModel internal constructor(
...
         }
 
-        fun mustShowFirstRun(
-            sharedPreferencesModel: SharedPreferencesModel,
-            loginToken: LoginToken
-        ): Boolean {
+        fun mustShowFirstRun(sharedPreferencesModel: SharedPreferencesModel, ): Boolean {
             return when {
-                // User has already signed in.
-                loginToken.isNotEmpty() -> false
```
Kokobe, line 119: @dan-alcantara-neeva anticipating your possible concern with this:

Since you are able to login from the WelcomeFlowActivity (which is now the new FirstRun) but not finish the WelcomeFlowActivity, you can get into a state where `finishWelcomeFlow()` is never called but `firstRunModel.mustShowFirstRun()` returns `false`. 

An example path of how to get here was:
1. Start the Neeva app for the first time
2. Neeva App: Opens Welcome Screen
3. Press "I have an account button"
4.  Neeva App: Open Sign-in Screen and Sign in
5.  Neeva App: redirects you to the Set Default Browser Screen
6. (at this point, firstRunModel.mustShowFirstRun() is wrongly returning false because the loginToken is not empty)
7.  Press back all the way to the Sign in Screen 
8. Sign in again
9. Neeva App: Brings you to ZeroQuery because after a sign-in & mustShowFirstRun() == false (we actually expect it to open the Set Default Browser Screen. 

The definition of `mustShowFirstRun()` needed to be changed because you can now login in FirstRun.

app/src/main/java/com/neeva/app/logging/ClientLogger.kt
```diff
@@ -90,7 +90,7 @@ class ClientLogger(
             return
         }
 
-        if (FirstRunModel.mustShowFirstRun(sharedPreferencesModel, loginToken)) {
+        if (loginToken.isNotEmpty() && !FirstRunModel.mustShowFirstRun(sharedPreferencesModel)) {
```
dan-alcantara-neeva, line 93: Talked over Zoom, but I think this conditional is also wrong.

* `addPendingLog` is a call we make to queue up logs to send when the user has finished first run.  We actively avoid sending anything until we're sure they've had a chance to say that they're OK with us sending logs, which means that they've gotten through the first screen of the flow.

* Because of the way SharedPreferences are _not_ set when we try to read them, we can't be certain that older users have seen first run or not.  We used "the user is signed in" as a substitute signal for that.

* This conditional is backwards from what it used to do, which means that we won't be sending any logs at all if the user has signed in && have completed first run, and will instead be sending logs (possibly without permission) in every other situation.

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -108,7 +109,7 @@ class WelcomeFlowActivity : AppCompatActivity() {
 
     private var sendUserToBrowserOnResume: Boolean = false
     private var selectedSubscriptionPlanTag: String? = null
-    private var loginReturnParams: LoginReturnParams? = null
+    private var initialLoginParams = mutableStateOf<LoginReturnParams?>(null)
```
dan-alcantara-neeva, line 112: 1. `mutableState` is a Compose thing.
2. Do you need a `mutableState` that is observable?  You seem to be using it as a boolean var everywhere.

dan-alcantara-neeva (changes requested):

app/src/main/java/com/neeva/app/logging/ClientLogger.kt
Kokobe: Made a fix for this in `MainActivity`. See `startUpWithoutLogin()`. 

When you open the app without a login intent, it will run the following conditional which decides which activity to open:

```
val activityClass = if (loginToken.isEmpty() && firstRunModel.mustShowFirstRun()) {
            WelcomeFlowActivity::class.java
        } else {
            firstRunModel.setFirstRunDone()
            NeevaActivity::class.java
        }
```

Kokobe: Any user that is already logged in will skip the WelcomeFlowActivity and have the sharedPref set that FirstRun is done. 

Users that are NOT logged in but have `FirstRunDone` == false, will see the WelcomeFlowActivity... and eventually have the FirstRunDone set to true in their shared preferences.

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
Kokobe: I need `mutableState` to cause the composables that use `initialLoginParams` (see `CreateAccountScreen`) to recompose and change what screen/activity to return to after login.

The reason we need a recomposition is if an old WelcomeFlowActivity is reused, the CreateAccountScreen screen Compose is showing is also old and has the incorrect `initialLoginParams` value. When that value changes (due to a `onNewIntent()` call), the UI will correctly recompose with the right loginReturnParams

Kokobe: Though it might not be a problem anymore because we set the WelcomeFlowActivity to `android:excludeFromRecents="true"`. This is more just "defensive coding" to protect for old state messing with the current state.

app/src/main/java/com/neeva/app/MainActivity.kt
```diff
@@ -53,14 +56,15 @@ class MainActivity : AppCompatActivity() {
...
             // outdated shared preference values and launch the app normally.
             firstRunModel.clearDestinationsToReturnAfterLogin()
-            launchNeevaActivity()
+            startUpWithoutLogin()
         }
     }
 
-    private fun launchNeevaActivity() {
-        val activityClass = if (firstRunModel.mustShowFirstRun()) {
+    private fun startUpWithoutLogin() {
```
dan-alcantara-neeva, line 63: `startUpWithoutLogin` sounds like you're trying to start the app up without logging in.  That codepath would also be triggered if you were logged in already, or if there was no login intent.

app/src/main/java/com/neeva/app/MainActivity.kt
```diff
@@ -117,6 +122,8 @@ class MainActivity : AppCompatActivity() {
         )
         firstRunModel.queueFetchNeevaInfo()
 
+        removeSingleTabActivityFromRecents()
```
dan-alcantara-neeva, line 125: Explain why you're doing this here.

app/src/main/java/com/neeva/app/MainActivity.kt
```diff
@@ -130,6 +137,20 @@ class MainActivity : AppCompatActivity() {
...
+    private fun removeSingleTabActivityFromRecents() {
+        try {
+            val am = this.getSystemService(Context.ACTIVITY_SERVICE) as ActivityManager?
+            val appTasks = am?.appTasks
+            appTasks
+                ?.filter {
+                    it.taskInfo.baseActivity?.className == SingleTabActivity::class.java.name
+                }
+                ?.forEach { it.finishAndRemoveTask() }
+        } catch (e: java.lang.Exception) {
```
dan-alcantara-neeva, line 149: Be more descriptive, or don't log it at all.  Also, you might be importing the wrong `Exception` class.

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
```diff
@@ -203,6 +197,7 @@ class FirstRunModel internal constructor(
             .appendQueryParameter("finalPath", provider.finalPath)
             .appendQueryParameter("signup", signup.toString())
             .appendQueryParameter("ignoreCountryCode", "true")
+            // TODO(kobec not yusuf):  mktEmailOutput https://github.com/neevaco/neeva-ios/blob/a714a70058dcedf4b5d019f8f097326c7bbc9894/Shared/NeevaConstants.swift#L270-L288
```
dan-alcantara-neeva, line 200: File issue or track it

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
```diff
@@ -411,7 +406,7 @@ class FirstRunModel internal constructor(
         }
 
         launchLoginFlow(
-            context,
+            context.applicationContext,
```
dan-alcantara-neeva, line 409: Say that you're doing this to force the `SingleTabActivity` to show up in its own task, and why that's important.

app/src/main/java/com/neeva/app/userdata/SessionToken.kt
```diff
@@ -98,12 +98,22 @@ abstract class SessionToken(
...
+                    // the cached cookie is equal to the Browser's cookie, the callback would
+                    // return false even if the Browser's cookie is up to date.
                     updateCookieManager(restoredValue) {
-                        continuation.resume(it)
+                        continuation.resume(Unit)
                     }
                 }
-                if (!wasCookieJarUpdated) {
+
+                // To actually tell the browser cookie jar is up to date, we have to ask the
```
dan-alcantara-neeva, line 111: tell if the

app/src/main/java/com/neeva/app/userdata/SessionToken.kt
```diff
@@ -98,12 +98,22 @@ abstract class SessionToken(
...
+                // browser for our cookie:
+                val (isBrowserAlive, cookiePair) = weakBrowser.getCookieFromBrowser()
+                val currentCookieValue = cookiePair?.value ?: ""
+
+                if (isBrowserAlive && currentCookieValue != cachedValue) {
```
dan-alcantara-neeva, line 116: Feels like this should be an or instead... what gets passed in for `onFailure`?

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -137,6 +136,8 @@ class WelcomeFlowActivity : AppCompatActivity() {
             clientLogger = clientLogger
```
dan-alcantara-neeva, line 136: Can you move `setDefaultAndroidBrowserManager = blah blah` earlier?  I don't think you actually need to hold off on putting it this low, and you might again get into a scenario it isn't set on time.

app/src/main/java/com/neeva/app/MainActivity.kt
Kokobe: How does `startUpWithoutLoginIntent` sound?

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
Kokobe: I'll do it in the next commit actually

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
Kokobe: Do you remember why it is important anymore? 

I originally thought it was so that `onNewIntent()` will be called whenever we launch a `SingleTabActivity`, but it has the same exact behavior (both call onNewIntent() appropriately) with or without the application context.

dan-alcantara-neeva: Yeah, we changed the AndroidManifest to force that behavior instead.  It's up to you to keep it or not

app/src/main/java/com/neeva/app/userdata/SessionToken.kt
Kokobe: Ah i assumed that if the browser wasn't alive then the cookie can't have possibly been set. 

Totally wrong, I should just check the current cookie value

Kokobe: onFailure() is just the code that says `popupModel.showSnackbar(getString(R.string.error_generic))`

app/src/main/java/com/neeva/app/appnav/AppNavModel.kt
```diff
@@ -48,7 +48,7 @@ interface AppNavModel {
     fun showShareSpaceSheet(spaceId: String)
     fun showSignInFlow()
 
-    fun showWelcomeFlow(loginReturnParams: LoginReturnParams)
+    fun showWelcomeFlow(loginReturnParams: LoginReturnParams, signInOnly: Boolean)
```
dan-alcantara-neeva, line 51: Clarify what `signInOnly` means here so that callers don't accidentally assume that it means something else.

app/src/main/java/com/neeva/app/settings/defaultbrowser/SetDefaultAndroidBrowserManager.kt
```diff
@@ -26,7 +26,7 @@ abstract class SetDefaultAndroidBrowserManager {
 
     abstract fun isRoleManagerAvailable(): Boolean
 
-    abstract fun requestToBeDefaultBrowser(callback: () -> Unit)
+    abstract fun requestToBeDefaultBrowser(callback: (Boolean) -> Unit)
```
dan-alcantara-neeva, line 29: Can you name the boolean in this callback?

app/src/main/java/com/neeva/app/settings/defaultbrowser/SetDefaultAndroidBrowserPane.kt
```diff
@@ -68,7 +68,7 @@ fun SetDefaultAndroidBrowserPane(
     openAndroidDefaultBrowserSettings: () -> Unit,
     setDefaultAndroidBrowserManager: SetDefaultAndroidBrowserManager,
     showAsDialog: Boolean,
-    onActivityResultCallback: () -> Unit,
+    onActivityResultCallback: (Boolean) -> Unit,
```
dan-alcantara-neeva, line 71: name the boolean if possible

app/src/main/java/com/neeva/app/welcomeflow/PlansScreen.kt
```diff
@@ -56,15 +56,31 @@ import timber.log.Timber
...
         .collectAsState().value
         ?.subscriptionOfferDetails
 
+    val initialTabIndex = if (showFreePlan) {
+        1
+    } else {
+        0
+    }
+
+    val onBackToUse = if (showBackButton) {
```
dan-alcantara-neeva, line 78: This API is a little weird... can you pass `onBack` down to `PlansScreen` and then tell it whether or not `showBackButton` is true or false?

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -216,78 +232,207 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
                                         this@WelcomeFlowActivity,
                                         selectedSubscriptionPlanTag
                                     )
                                 },
-                                navigateToSignIn = loginFlowNavModel::navigateToSignIn,
-                                onBack = { navHost.popBackStack() },
+                                navigateToSignIn = welcomeFlowNavModel::showSignIn,
+                                onBack = onBack,
                             )
+                            LaunchedEffect(true) {
```
dan-alcantara-neeva, line 286: newline above

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -216,78 +232,207 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
                                 onPremiumAvailable = {
                                     subscriptionManager.buy(
                                         this@WelcomeFlowActivity,
                                         selectedSubscriptionPlanTag
                                     )
                                 },
                                 navigateToCreateAccount =
-                                loginFlowNavModel::navigateToCreateAccountWithGoogle,
-                                onBack = { navHost.popBackStack() },
+                                welcomeFlowNavModel::showCreateAccountWithGoogle,
```
dan-alcantara-neeva, line 309: use braces to avoid the wonky indentation

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -216,78 +232,207 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
+    private fun setupStartScreen() {
+        val screenToReturnToAfterLogin = firstRunModel.getScreenToReturnToAfterLogin()
+        firstRunModel.clearDestinationsToReturnAfterLogin()
+
+        val loginHappened = !neevaUser.isSignedOut()
+
+        when {
+            // If there was a login, navigate to the specified screen.
+            loginHappened && isValidDestination(screenToReturnToAfterLogin) -> {
+                if (this::navHost.isInitialized) {
```
dan-alcantara-neeva, line 398: Why do you need the `this::` in so many places?

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -341,32 +476,18 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
-            val isNewUser = firstRunModel.mustShowFirstRun()
-            val optedForPremium = selectedSubscriptionPlanTag == ANNUAL_PREMIUM_PLAN ||
-                selectedSubscriptionPlanTag == MONTHLY_PREMIUM_PLAN
-
-            if (isNewUser || optedForPremium) {
-                return LoginReturnParams(
-                    activityToReturnTo = WelcomeFlowActivity::class.java.name,
-                    screenToReturnTo = Destinations.SET_DEFAULT_BROWSER.name
-                )
+    private fun getOnContinueInPlansScreen(): () -> Unit = {
```
dan-alcantara-neeva, line 479: What's with the equals sign in this function definition?

app/src/main/java/com/neeva/app/welcomeflow/WelcomeScreen.kt
```diff
@@ -133,21 +143,20 @@ internal fun ConsentCheckbox(
...
-        ?.subscriptionOfferDetails
-
     WelcomeFlowStackedButtons(
         primaryText = stringResource(id = R.string.welcomeflow_lets_go),
         onPrimaryButton = {
-            if (offers.isNullOrEmpty()) {
-                navigateToSetDefaultBrowser()
-            } else {
-                navigateToPlans()
+            when {
```
dan-alcantara-neeva, line 156: 1. Was this code duplicated somewhere else? 
2. Pass in a single lambda that calls the correct function instead of passing in all the lambdas and pushing the logic down here

app/src/main/java/com/neeva/app/appnav/AppNavModel.kt
Kokobe: Okay, but will remove the comment in the next PR 

Follow-up PR is going to do add a `Purpose` where it should be clear what we are doing with `showWelcomeFlow()` : https://github.com/neevaco/neeva-android/blob/939b3a1c9edaf18b2c485589120287f102eaf717/app/src/main/java/com/neeva/app/appnav/AppNavModel.kt

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
Kokobe: The only way to safely check if a `lateinit` has been initialized is to use `this::[variable name].isInitialized`

see: https://stackoverflow.com/questions/37618738/how-to-check-if-a-lateinit-variable-has-been-initialized

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
Kokobe: Bad syntax by me

app/src/main/java/com/neeva/app/welcomeflow/WelcomeScreen.kt
Kokobe: Agh great advice! thank you

app/src/main/java/com/neeva/app/firstrun/widgets/texts/EmailPromoCheckbox.kt
```diff
@@ -15,45 +15,39 @@ import androidx.compose.material3.CheckboxDefaults
...
 @Composable
-fun EmailPromoCheckbox(
-    mktEmailOptOut: Boolean,
-    setMktEmailOptOut: (Boolean) -> Unit,
-    modifier: Modifier = Modifier
-) {
+fun EmailPromoCheckbox(modifier: Modifier = Modifier) {
+    val firstRunModel = LocalFirstRunModel.current
+    val toggleMktEmailOptOut: (Boolean) -> Unit = {
+        firstRunModel.mktEmailOptOutState.value = !firstRunModel.mktEmailOptOutState.value
```
dan-alcantara-neeva, line 33: This code should be in the model itself.  You're pulling the value of a boolean out and then setting it back in the same place

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -180,29 +180,24 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
                         composable(Destinations.PLANS.name) {
                             PlansScreen(
-                                onContinue = getOnContinueInPlansScreen(),
+                                onContinue = ::onContinueInPlansScreen,
                                 navigateToSignIn = welcomeFlowNavModel::showSignIn,
                                 saveSubscriptionPlanChoice = { tag ->
                                     selectedSubscriptionPlanTag = tag
                                 },
-                                onBack = onBack,
+                                onBack = if (!firstRunModel.mustShowFirstRun()) {
```
dan-alcantara-neeva, line 195: Maybe use `takeIf`?  Don't know if it's more or less readable

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -476,7 +464,17 @@ class WelcomeFlowActivity : AppCompatActivity() {
         super.onSaveInstanceState(outState, outPersistentState)
     }
 
-    private fun getOnContinueInPlansScreen(): () -> Unit = {
+    private fun onContinueInWelcomeScreen() {
+        when {
+            isPremiumPurchaseAvailable() -> welcomeFlowNavModel.showPlans()
```
dan-alcantara-neeva, line 469: Be stylistically consistent with `onContinueInPlansScreen`: Use brackets and code blocks

dan-alcantara-neeva (approved):

PR #1199, opened 2023-01-14
Premium fix Billing API loading time + add checks before buying Premium

### LINKED ISSUE #1198: Include obfuscated id flow as a isBillingAvailable() parameter
See title

Kokobe:
The timing and use of fetching the `ProductDetails` and `ExistingPurchases` of a user from the Google Billing API was wrong. This made the Plans screen take a while to load its Subscription Offering data. 

This PR: 
* Reduces the loading time for fetching subscription offering data. 
* "Learn more about Premium" button in `ProfileSettingsPane` now leads the user to the neeva.com/settings/membership website in a browser tab if the Billing API or Premium is unavailable for them to purchase on this device. 
* Added flows and verbal logs for when Google Billing API and Neeva Premium Purchase ready.
* Adds a `ATTEMPTS_TO_ASSUME_FAILURE` to `BillingClientController` so that the user doesn't have to wait for the `MAX_ATTEMPTS` to finish. 
* Introduces a 1 minute timeout in case the user's internet connection is too slow and we can't fetch Billing information. 
* Adds a Loading indicator and disables the "Let's go" button until we get enough information to determine if the user can purchase Premium or not.
* Cleans up code in ActivityStarter

* https://www.figma.com/file/dylvDPB69JG9U1rzZ3HQ52/Android-Premium?node-id=0%3A1
* 1 min timeout for Welcome Screen conversation: https://neevaco.slack.com/archives/C03PQUPJJDP/p1674001779236809

### How was this tested?
- [ ] I added automated tests

# Manual test cases

## ProfilePane Changes
ProfilePane + press "Learn more about Premium" + billing and premium available -> Sends user to Plans screen.

https://user-images.githubusercontent.com/20916043/213127187-621d9681-a119-4aad-a5cf-f63aba7b2c01.mp4

ProfilePane + press "Learn more about Premium" + billing and premium not available -> Sends user to neeva.com/settings/membership

https://user-images.githubusercontent.com/20916043/212504449-348072a2-dd32-4069-b133-3e364ef64e93.mp4

## Welcome Screen Lets Go button disabling

FirstRun Billing disabled -- no effect on Sign-in because the toggle will cause the `isBillingAvailableFlow` to immediately return false:

https://user-images.githubusercontent.com/20916043/212504469-14fc6af3-ab47-4134-b08e-2dea20827aa6.mp4

FirstRun Billing enabled -> let's go button is disabled until Billing API ready

https://user-images.githubusercontent.com/20916043/213129511-8a58748c-6133-4cdc-9179-a302a0c0e951.mp4

FirstRun but Billing never gets back to us (1 min of waiting time before we allow the user to move on)

https://user-images.githubusercontent.com/20916043/213098592-1513c827-fce8-4d4e-b558-27727be2bc65.mov


app/src/test/java/com/neeva/app/userdata/NeevaUserTest.kt
```diff
@@ -93,7 +93,7 @@ class NeevaUserTest : BaseTest() {
             sharedPreferencesModel = sharedPreferencesModel,
             loginToken = loginToken,
             networkHandler = networkHandler,
-            billingClientController = billingClientController
+            billingClientController = billingClientController,
```
dan-alcantara-neeva, line 96: Unnecessary comma changes; revert all of these to make the PR cleaner

app/src/main/java/com/neeva/app/welcomeflow/WelcomeScreen.kt
```diff
@@ -124,12 +127,17 @@ internal fun ConsentCheckbox(
 
 @Composable
 private fun ContinueButtons(
-    onContinueInWelcomeScreen: () -> Unit,
+    onContinueInWelcomeScreen: (() -> Unit)?,
     navigateToSignIn: () -> Unit,
 ) {
+    var onLetsGo = { }
```
dan-alcantara-neeva, line 133: Can you just use a regular if statement here instead of this funky `?.let` block?

app/src/main/java/com/neeva/app/settings/SettingsController.kt
```diff
@@ -195,13 +199,22 @@ class SettingsControllerImpl(
...
-                appNavModel.showWelcomeFlow(
-                    LoginReturnParams(
-                        activityToReturnTo = NeevaActivity::class.java.name,
-                        screenToReturnTo = AppNavDestination.SETTINGS.name
-                    ),
-                    purpose = WelcomeFlowActivity.Companion.Purpose.BROWSE_PLANS
-                )
+                val canPurchasePremium = subscriptionManager.isBillingAvailableFlow.value == true &&
+                    subscriptionManager.isPremiumPurchaseAvailableFlow.value == true
+                when (canPurchasePremium) {
```
dan-alcantara-neeva, line 204: This is just an if statement.

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -177,7 +178,7 @@ class WelcomeFlowActivity : AppCompatActivity() {
                     ) {
                         composable(Destinations.WELCOME.name) {
                             WelcomeScreen(
-                                onContinueInWelcomeScreen = ::onContinueInWelcomeScreen,
+                                onContinueInWelcomeScreen = onContinueInWelcomeScreen(),
```
dan-alcantara-neeva, line 181: This function name and invocation are misleading: you're calling a function to create another function that is actually called when the user hits the button.  Instead of doing that, can you just call the function like you used to and make that do the correct checks?

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -474,16 +469,27 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
-            isPremiumPurchaseAvailable() -> {
-                welcomeFlowNavModel.showPlans()
-            }
-            !setDefaultAndroidBrowserManager.isNeevaTheDefaultBrowser() -> {
-                welcomeFlowNavModel.showSetDefaultBrowser()
-            }
-            else -> {
-                finishWelcomeFlow()
+    @Composable
+    private fun onContinueInWelcomeScreen(): (() -> Unit)? {
```
dan-alcantara-neeva, line 473: Why is this a Composable?

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -39,40 +45,78 @@ class SubscriptionManager(
...
         get() = SharedPrefFolder.FirstRun.SelectedSubscriptionTag.getFlow(sharedPreferencesModel)
 
-    val productDetailsFlow = billingClientController.productDetailsFlow
-    val purchasesFlow = billingClientController.purchasesFlow
-    val obfuscatedUserIDFlow = billingClientController.obfuscatedUserIDFlow
+    val productDetailsWrapperFlow = billingClientController.productDetailsFlowWrapper
+    val existingPurchasesFlow = billingClientController.purchasesFlow
+    val isBillingAvailableFlow = productDetailsWrapperFlow
+        .combine(existingPurchasesFlow) { productDetailsWrapper, existingPurchases ->
+            if (!settingsDataModel.getSettingsToggleValue(SettingsToggle.DEBUG_ENABLE_BILLING))
```
dan-alcantara-neeva, line 60: Always put braces when it ends up on multiple lines

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -39,40 +45,78 @@ class SubscriptionManager(
...
+                return@combine false
 
-    val hasAnnualPremiumPlan: Flow<Boolean> = billingClientController.purchasesFlow
-        .map { purchaseList ->
-            purchaseList.any { purchase ->
-                purchase.products.contains(BillingSubscriptionPlanTags.ANNUAL_PREMIUM_PLAN)
+            // If the Billing information has not been fetched yet, then we still don't know if
+            // Billing is available for this user.
+            if (!productDetailsWrapper.isSet || existingPurchases == null) {
+                Log.v(
```
dan-alcantara-neeva, line 66: Timber, not Log.

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
Kokobe: I wanted to make the WelcomeScreenComposable re-compose itself whenever the `isBillingAvailableFlow` changed see: `val isBillingAvailable = subscriptionManager.isBillingAvailableFlow.collectAsState().value`

I can pull that out though and make this function not a Composable

app/src/main/java/com/neeva/app/billing/billingclient/FetchPurchasesManager.kt
```diff
@@ -53,9 +54,12 @@ class FetchPurchasesManager(
                     _purchasesFlow.value = purchaseList
                 } else {
                     fetchContinuation.resume(false)
+                    if (shouldAssumeFailureForNow) {
```
dan-alcantara-neeva, line 57: What is `shouldAssumeFailureForNow` supposed to mean?  Add it to the kdoc.

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
dan-alcantara-neeva: It's a fragile link, regardless.  If you want to force something somewhere else to recompose when necessary, make the Composable have the logic for knowing when to trigger it -- not a callback that a button on that screen displays.

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
Kokobe: But Timber doesn't work on release builds? I guess I can look into it and add it to the growing scope of Premium

app/src/main/java/com/neeva/app/welcomeflow/WelcomeScreen.kt
Kokobe: Good point lol

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -146,6 +190,16 @@ class SubscriptionManager(
             }
     }
 
+    private fun collectIsPremiumPurchaseAvailable(onResult: (Boolean) -> Unit) {
+        coroutineScope.launch(dispatchers.io) {
+            isPremiumPurchaseAvailableFlow.collect {
```
dan-alcantara-neeva, line 195: Do you need this to stay alive after the first `it != null` condition is hit?  Can you make the job terminate itself so that it's not just constantly running?  There should be operators other than `collect` that would help you do that...

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -168,26 +222,41 @@ class SubscriptionManager(
...
-        val offerToken = leastPricedOfferToken(offerDetails)
-        val obfuscatedUserID = billingClientController.obfuscatedUserIDFlow.value
-
-        if (isPremiumPurchaseAvailable()) {
-            offerToken?.let {
-                val billingParams = billingFlowParamsBuilder(
-                    productDetails = productDetails,
-                    offerToken = it,
-                    obfuscatedUserID = obfuscatedUserID
+        coroutineScope.launch(dispatchers.io) {
```
dan-alcantara-neeva, line 225: You've introduced a Job that will run whenever `buy` is called.  Can you guarantee that `buy()` will only called once?  If not, you will end up with two Jobs running in parallel.

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -238,5 +307,6 @@ class SubscriptionManager(
 
     companion object {
         val QUEUE_BUY_ON_SIGN_IN_JOB_NAME = "WelcomeFlow: QueueBuyOnSuccessfulSignIn"
+        val TAG = "SubscriptionManager"
```
dan-alcantara-neeva, line 310: No more tags; we're using Timber.

app/src/main/java/com/neeva/app/welcomeflow/WelcomeScreen.kt
dan-alcantara-neeva: Behavior is more of a UX question.  If the user is offline when they start the app, will this button ever be enabled?

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
Kokobe: I need to pass in a `isBillingAvailable` here because `onContinueInWelcomeScreen` depends on that variable. I'll rename the function to: `getOnContinueInWelcomeScreen()`

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
Kokobe: This if statement is only 1 line, but I'll put braces around it anyways and add a log here

dan-alcantara-neeva: If the return ends up on the next line, you have to add braces.

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
Kokobe: Is calling `this@launch.cancel()` after `onResult(it)` happens an adequate way to do this? 

It will cancel the job created by `launch`, which will end the collecting when the `it != null` condition is hit

dan-alcantara-neeva: Is there a `single` or something similar?  Might need to combine it with a filter

app/src/main/java/com/neeva/app/welcomeflow/WelcomeScreen.kt
Kokobe: Yes this button will be enabled in 1.3 seconds because of my `ATTEMPTS_TO_ASSUME_FAILURE` in `BillingClientController`.

Basically, if we the fetch ProductDetails job fails twice, we tell the the UI to assume the job failed and allow the user to move on to the next screen (not PLANS related). This doesn't mean we stop trying to fetch it (it's still happening with exponential backoff), but we allow the user to continue on. 

You can see this in my "FirstRun billing disabled" video in the PR description

dan-alcantara-neeva: Are you just making the button do nothing by removing the onClick?  Can you properly mark the button as disabled so that it doesn't stay blue and look clickable?

app/src/main/java/com/neeva/app/billing/billingclient/FetchPurchasesManager.kt
Kokobe: I've added documentation in `BillingClientController`'s `ATTEMPTS_TO_ASSUME_FAILURE` constant which is where the variable originates and is ultimately passed onto `FetchProductDetailsManager` and `FetchPurchasesManager`

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
Kokobe: From what I'm reading it seems that `single` is still a `collect` that is constantly running. Why is single better? 

Screen Shot 2023-01-17 at 2 48 57 PM

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
dan-alcantara-neeva: We shouldn't be mixing and matching Timber and Log usage; we should be sticking with a single framework for logging our messages.  If Timber isn't good enough for what we need, we should rip it back out.

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
Kokobe: Timber is still not working on release

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -36,43 +43,92 @@ class SubscriptionManager(
...
+    val productDetailsWrapperFlow = billingClientController.productDetailsFlowWrapper
+    val existingPurchasesFlow = billingClientController.purchasesFlow
+    val isBillingAvailableFlow = productDetailsWrapperFlow
+        .combine(existingPurchasesFlow) { productDetailsWrapper, existingPurchases ->
+            if (!settingsDataModel.getSettingsToggleValue(SettingsToggle.DEBUG_ENABLE_BILLING)) {
+                Log.v(
+                    TAG,
+                    "Billing not available because the dev `Enable billing` settings toggle " +
+                            "is disabled."
+                    )
```
dan-alcantara-neeva, line 71: ktlint moved this parenthesis here?

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
dan-alcantara-neeva: Talked offline; it's fine to use Log for now because Timber isn't doing what we need in release builds.  Will revisit once the rest of the flow is in place.

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -141,11 +197,23 @@ class SubscriptionManager(
...
             }
     }
 
+    private fun collectIsPremiumPurchaseAvailable(onResult: (Boolean) -> Unit) {
+        buySubscriptionCoroutineScope.launch(dispatchers.io) {
+            isPremiumPurchaseAvailableFlow
+                .collect {
+                    Log.e("KOBE", "Collected Premium Purchase available")
+                    if (it == true) {
+                        onResult(true)
```
dan-alcantara-neeva, line 211: Still need this Job to be canceled so that it doesn't keep firing your callback

app/src/main/java/com/neeva/app/welcomeflow/WelcomeScreen.kt
Kokobe: Lol! No but it wasn't clear since that logic was happening a level lower... I can move it up

Kokobe:
Updated the videos @dan-alcantara-neeva

app/src/main/res/values/strings.xml
```diff
@@ -1029,6 +1029,9 @@
     <!-- [WelcomeFlow] Text for Settings Row to learn more about the Premium subscription. -->
     <string name="welcomeflow_learn_more_about_premium">Learn more about Premium</string>
 
+    <!-- [WelcomeFlow] In the Welcome Screen, this text tells the user that we are loading and we are not ready to show them the next Screen yet. Replaces the "Let's go" Button's text label while we are loading. -->
+    <string name="welcomeflow_loading">Loading</string>
```
dan-alcantara-neeva, line 1033: Just make this a generic string called "loading" so that we can reuse it.

app/src/main/java/com/neeva/app/appnav/ActivityStarter.kt
```diff
@@ -17,6 +18,18 @@ import timber.log.Timber
 
 class ActivityStarter(private val appContext: Context, private val popupModel: PopupModel) {
 
+    fun safeStartActivityWithoutActivityTransition(intent: Intent) {
+        safeStartActivityForIntent(
+            intent = intent,
+            // Nullify the transition animation to hide the fact that we're switching Activities.
+            options = ActivityOptionsCompat.makeCustomAnimation(
```
dan-alcantara-neeva, line 25: Just put this on one line; we can't even name what each argument is because it's a Java call.

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -223,12 +221,11 @@ class SubscriptionManager(
         tag: String?,
         onBillingFlowFinished: (BillingResult) -> Unit
     ) {
-        buySubscriptionCoroutineScope.cancel()
-        buySubscriptionCoroutineScope = CoroutineScope(Job() + dispatchers.main)
+        cancelPurchaseJobsAndResetScope()
```
dan-alcantara-neeva, line 224: Are you down to one `Job` being launched and the `coroutineScope` being used in one place?  Can you just hold onto a `Job?` instead of creating another CoroutineScope?  You can call `cancel()` on it when you're done and create a new one when needed.

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
```diff
@@ -194,4 +188,17 @@ class BillingClientController(
...
+    companion object {
+        // region Retrying BillingClient connection parameters
+        internal val SLOW_CONNECTION_WAIT_TIME = 1000L * 60
+        internal val MAX_ATTEMPTS = 5
+        private val FIRST_RETRY_WAIT_TIME = 300L
+        private val RETRY_WAIT_TIME_MULTIPLIER = 2.0
+        // If a job fails this many times, we will assume failure for now. We will continue to try the
+        // job in the background until the max attempts have been finished.
+        private val ATTEMPTS_TO_ASSUME_FAILURE = 4
+        private var attempts = 0
```
dan-alcantara-neeva, line 201: Why is this on the companion object?

app/src/main/java/com/neeva/app/billing/billingclient/FetchPurchasesManager.kt
```diff
@@ -32,6 +35,15 @@ class FetchPurchasesManager(
 
     private var callback: (BillingResult) -> Unit = { }
 
+    private val slowInternetConnectionTimer: Job
+
+    init {
+        slowInternetConnectionTimer = coroutineScope.launch(dispatchers.io) {
```
dan-alcantara-neeva, line 41: This should be launched at the same time as the `fetchPurchases` call because you're timing that -- not when this class was first initialized.

app/src/main/java/com/neeva/app/welcomeflow/SharedComposables.kt
```diff
@@ -32,26 +34,40 @@ internal fun WelcomeFlowButton(
...
-            modifier = Modifier.padding(vertical = Dimensions.PADDING_SMALL)
-        ) {
-            Text(
-                text = primaryText,
-                style = MaterialTheme.typography.titleMedium
-            )
-            if (secondaryText != null) {
+        Row(verticalAlignment = Alignment.CenterVertically) {
+            if (!enabled) {
+                CircularProgressIndicator(
```
dan-alcantara-neeva, line 47: 1. It's a small thing, but Hwi wanted this to show up only after 300 milliseconds.
2. Also a small thing, but you could use a `Crossfade` composable that's dependent on the `enabled` flag as its `targetState` so that you could show the button fading as it goes from being disabled to enabled.

dan-alcantara-neeva (changes requested):

app/src/main/java/com/neeva/app/billing/billingclient/BillingClientController.kt
Kokobe: moved it into the class, you're right only constants should be in here

app/src/main/java/com/neeva/app/welcomeflow/SharedComposables.kt
Kokobe: Added this to a list of Polish action items, will punt on this issue for the next PR

dan-alcantara-neeva (approved):

app/src/main/java/com/neeva/app/billing/billingclient/FetchProductDetailsManager.kt
```diff
@@ -28,8 +43,18 @@ class FetchProductDetailsManager(private val billingClientController: BillingCli
                     .build()
             )
         )
+    private val slowInternetConnectionTimer: Job
 
-    suspend fun fetchProductDetails(): Boolean = suspendCoroutine { fetchContinuation ->
+    init {
+        slowInternetConnectionTimer = coroutineScope.launch(dispatchers.io) {
```
dan-alcantara-neeva, line 49: I think I asked you to move this offline; we want this to start when the job does and not when the class is initialized.

Kokobe: Fixed this in #1235

PR #1209, opened 2023-01-19
Adjust timeout so that it happens during fetch instead init

### LINKED ISSUE #1208: `FetchProductDetailsManager` should start the cancelation timeout only after `fetchProductDetails` is called
The way it's currently set up `init`, the timer will start immediately after the class is initialized, which could be a while before the call it's supposed to be monitoring is actually fired.

Kokobe:
Adjusts `slowInternetConnectionTimeout` so that it happens during fetch instead init in `fetchProductDetails` and `fetchPurchases`. 

### How was this tested?
- [ ] I added automated tests

## Issues addressed
* Fixes #1208


app/src/main/java/com/neeva/app/billing/billingclient/FetchProductDetailsManager.kt
```diff
@@ -43,18 +43,16 @@ class FetchProductDetailsManager(
...
 
-    init {
-        slowInternetConnectionTimer = coroutineScope.launch(dispatchers.io) {
+    suspend fun fetchProductDetails(
+        shouldAssumeFailureForNow: Boolean
+    ): Boolean = suspendCoroutine { fetchContinuation ->
+        slowInternetConnectionTimeout = coroutineScope.launch(dispatchers.io) {
             delay(SLOW_CONNECTION_WAIT_TIME)
             assumeFailureToFetch()
         }
```
dan-alcantara-neeva, line 54: Can you add a log between line 51 and 52, then another log right before line 56?  Want to make sure this job is launched asynchronously and doesn't block the rest of the function

Kokobe: Confirmed that it doesn't block the rest of the function

dan-alcantara-neeva (approved):

PR #1221, opened 2023-01-23
Add google native sign navigation logic + Loading Screen

Kokobe:
This PR adds a UI reaction to the Google Native Sign in. This includes:
* Adding a `LoadingScreen` for when a Google Native Sign In occurs and we are fetching the user information. (LoadingScreen will be added for the other sign in options in the follow-up PR)
* Makes the `LoadingScreen` stay until we know if the user can purchase Premium or not. 
* Adds a `navigateToScreenAfterLoginWhenReady()` to `WelcomeFlowNavModel` 
  * Which means I have to move `finishWelcomeFlow()` to `WelcomeFlowNavModel`

* Loading screen figma: https://www.figma.com/file/dylvDPB69JG9U1rzZ3HQ52/Android-Premium?node-id=755%3A4576&t=cZsILUNWwxoYLL9w-0
* Loading screen slack convo: https://neevaco.slack.com/archives/C03PQUPJJDP/p1674281518990829

### How was this tested?
- [ ] I added automated tests

### Manual test cases
Cancelling a Native Google Sign in
Native Google Sign In (Happy Path)
Native Google Sign In (Slow Wifi)
Native Google Sign In (Fails but we open up the SingleTabActivity instead):

Cancelling a Native Google Sign in

https://user-images.githubusercontent.com/20916043/213867666-32309acd-150f-4a55-8a8f-4c2a9558c8bc.mp4

Native Google Sign In (Happy Path)

https://user-images.githubusercontent.com/20916043/213867704-008b67bb-c097-4eed-9a56-3632b5906ee1.mp4

Native Google Sign In (Slow Wifi) -- feel free to fast forward

https://user-images.githubusercontent.com/20916043/214083085-ab5824f4-c14d-4c56-bdf4-37b6b9ed1c92.mov

Native Google Sign In (Fails but we open up the SingleTabActivity instead):

https://user-images.githubusercontent.com/20916043/213872726-43444b3b-a99d-44cd-a367-3f335a143a1e.mp4


app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -202,9 +200,15 @@ class SubscriptionManager(
             }
     }
 
-    private suspend fun collectIsPremiumPurchaseAvailable(): Boolean {
+    suspend fun collectIsPremiumPurchaseAvailable(): Boolean {
         return isPremiumPurchaseAvailableFlow
-            .filter { it == true }
+            .filterNotNull()
+            .take(1)
```
dan-alcantara-neeva, line 206: `.take(1).single()` -> `.first()`, here and everywhere else

app/src/main/java/com/neeva/app/welcomeflow/login/LoadingScreen.kt
```diff
@@ -0,0 +1,60 @@
...
+import com.neeva.app.welcomeflow.WelcomeFlowContainer
+
+@Composable
+fun LoadingScreen() {
+    WelcomeFlowContainer {
+        Column(
+            horizontalAlignment = Alignment.CenterHorizontally,
+            modifier = it
+        ) {
+            Spacer(Modifier.size(249.dp))
```
dan-alcantara-neeva, line 31: Any time you're hard-coding a number like `249.dp`, I need to understand why you chose it and whether you should be centering the content & spinner, instead.  A hard-coded number will never work correctly across multiple screen sizes.

app/src/main/java/com/neeva/app/welcomeflow/login/LoadingScreen.kt
```diff
@@ -0,0 +1,60 @@
...
+
+@PortraitPreviews
+@Composable
+fun LoadingScreen_Light_Preview() {
+    NeevaThemePreviewContainer(useDarkTheme = false) {
+        LoadingScreen()
+    }
+}
+
+@PortraitPreviews
```
dan-alcantara-neeva, line 54: You need to use @PortraitPreviewsDark.

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowContainer.kt
```diff
@@ -117,14 +117,16 @@ fun WelcomeFlowHeader(
                     modifier = Modifier.width(89.dp)
                 )
                 Spacer(Modifier.height(Dimensions.PADDING_LARGE))
```
dan-alcantara-neeva, line 119: Should one of these spacers live inside of the `?.let` block?  Otherwise, you introduce an extra big space that unbalances the content.

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowContainer.kt
```diff
@@ -117,14 +117,16 @@ fun WelcomeFlowHeader(
...
-                    textAlign = TextAlign.Center,
-                    style = MaterialTheme.typography.titleMedium.copy(
-                        fontSize = 18.sp,
-                        fontWeight = FontWeight.Normal
+                headerText?.let {
+                    Text(
+                        text = it,
+                        textAlign = TextAlign.Center,
+                        style = MaterialTheme.typography.titleMedium.copy(
+                            fontSize = 18.sp,
```
dan-alcantara-neeva, line 125: If you're overriding the font size and font weight, what from the `titleMedium` are you actually using?  Can you ask if there's an equivalent font you should be using, instead?

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -217,52 +221,55 @@ class SubscriptionManager(
...
-        if (
-            tag != BillingSubscriptionPlanTags.MONTHLY_PREMIUM_PLAN &&
-            tag != BillingSubscriptionPlanTags.ANNUAL_PREMIUM_PLAN
-        ) {
-            return
-        }
-
         subscriptionJob = appCoroutineScope.launch(dispatchers.io) {
+            if (
+                tag != BillingSubscriptionPlanTags.MONTHLY_PREMIUM_PLAN &&
```
dan-alcantara-neeva, line 230: Define this condition somewhere and reuse it over and over again to avoid it getting out of sync with the other places you use the same check.

app/src/main/java/com/neeva/app/welcomeflow/login/WelcomeFlowNavModel.kt
```diff
@@ -34,25 +62,94 @@ class WelcomeFlowNavModel(private val navHost: NavHostController) {
...
+        } else {
+            clientLogger.logCounter(LogConfig.Interaction.SKIP_DEFAULT_BROWSER, null)
+        }
+        clientLogger.sendPendingLogs()
+
+        // An activity outside of WelcomeFlowActivity has triggered the WelcomeFlow via intent:
+        val returnToActivityAndScreen = destinationToReturnWhenFinished.value != null
+
+        activityReference.get()?.let { activity ->
+            val browserIntent = activity.intent.apply {
```
dan-alcantara-neeva, line 131: Can't `intent` be null?

app/src/main/java/com/neeva/app/welcomeflow/login/LaunchLoginFlow.kt
```diff
@@ -33,10 +36,15 @@ fun launchLoginFlow(
         ActivityResultContracts.StartActivityForResult()
     ) { activityResult ->
         firstRunModel.handleLoginActivityResult(context, activityResult, params) {
+            welcomeFlowNavModel.showLoadingScreen()
+            welcomeFlowNavModel.navigateToScreenAfterLoginWhenReady()
```
dan-alcantara-neeva, line 40: Move the `showLoadingScreen` call into `navigateToScrenAfterLoginWhenReady` call.  The way the function is named implies that it needs to do something while it's getting ready, which would be showing the loading screen.

app/src/main/java/com/neeva/app/ui/PreviewContainer.kt
```diff
@@ -211,7 +211,6 @@ fun PreviewCompositionLocals(content: @Composable () -> Unit) {
         settingsDataModel = previewSettingsDataModel,
         sharedPreferencesModel = previewSharedPreferencesModel
     )
-
```
dan-alcantara-neeva, line 214: Put this back.

app/src/main/java/com/neeva/app/welcomeflow/login/WelcomeFlowNavModel.kt
```diff
@@ -34,25 +62,94 @@ class WelcomeFlowNavModel(private val navHost: NavHostController) {
...
+            clientLogger.logCounter(LogConfig.Interaction.SET_DEFAULT_BROWSER, null)
+        } else {
+            clientLogger.logCounter(LogConfig.Interaction.SKIP_DEFAULT_BROWSER, null)
+        }
+        clientLogger.sendPendingLogs()
+
+        // An activity outside of WelcomeFlowActivity has triggered the WelcomeFlow via intent:
+        val returnToActivityAndScreen = destinationToReturnWhenFinished.value != null
+
+        activityReference.get()?.let { activity ->
```
dan-alcantara-neeva, line 130: 1. Why are you using an activity reference?  ViewModels should generally have no idea what Activity it's associated with because it's supposed to outlive the activity.  Can't you just use app context and pass the intent in?

2. Re-add `WelcomeFlowActivity#finishWelcomeFlow`, then make that call `WelcomeFlowNavModel.finishWelcomeFlow()` and `finishAndRemoveTask()`.  A `ViewModel` should generally avoid ever storing references to Views or Activities or anything that could be recreated in case the Activity dies.  It also means that you're violating the encapsulation between the View and Model.

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
```diff
@@ -146,15 +146,28 @@ class WelcomeFlowActivity : AppCompatActivity() {
...
+
         processIntent(intent)
         firstRunInitialization()
 
         setContent {
             NeevaTheme {
                 navHost = rememberAnimatedNavController()
-                welcomeFlowNavModel = remember(navHost) {
-                    WelcomeFlowNavModel(navHost)
+                LaunchedEffect(navHost) {
```
dan-alcantara-neeva, line 167: Make this a disposable effect.

```
DisposableEffect(navHost) {
    welcomeFlowNavModel.setNavHost(navHost)
    onDispose { welcomeFlowNavModel.setNavHost(null) }
}
```

Avoids holding onto dead classes.  Should be doing this in `NeevaActivity` too, honestly...

dan-alcantara-neeva (changes requested):
I'm worried that the current state of the login flow is too fragile and confusing.  Can you take a step back and write up how this is all supposed to work, including:

* How your activity/screen to return to/screen to return to after failure mechanisms work
* Describing the different login states that a user can be in
* Links between the different states the user can be in and how they transition between them
* Whatever else someone two months from now would need to understand how your code fits into everything?

Kokobe:
> I'm worried that the current state of the login flow is too fragile and confusing. Can you take a step back and write up how this is all supposed to work, including:
> 

> * Describing the different login states that a user can be in
> * Links between the different states the user can be in and how they transition between them
> * Whatever else someone two months from now would need to understand how your code fits into everything?

The follow-up PR #1223 clarifies what the final state of the login flow will look like.
Important points: 
* Any login will get redirected to the `LoadingScreen` in the `WelcomeFlowActivity`. 
* Any login failure will get redirected to the `SignInScreen` in the `WelcomeFlowActivity`
* All `activityToReturnTo` will be used for is returning the user to the correct activity after finishing the WelcomeFlow. 
* The enum `Purpose` defines the flow behavior + which starting screen in the WelcomeFlow

### As for the current state in this PR:
> * How your activity/screen to return to/screen to return to after failure mechanisms work

Any screen

app/src/main/java/com/neeva/app/welcomeflow/login/LoadingScreen.kt
Kokobe: Since we made the `WelcomeFlowContainer` scrollable, it is actually impossible to vertically center the content in there. In other terms: the content is allowed to grow as large as it wants, which makes the height of the container unknown and therefore we don't know the center (unless we recompose and do some screen sizing calculations after the content has been placed down). 

But the designs aren't centered anways:
https://user-images.githubusercontent.com/20916043/214201477-e43b7ddf-5a51-4334-bb72-62ec60117a17.mov

Kokobe: Should I create a separate `WelcomeFlowContainerNoScrollable` just for the `LoadingScreen`? I can do that as a follow-up PR

app/src/main/java/com/neeva/app/welcomeflow/login/LoadingScreen.kt
Kokobe: Sure! What is the difference? I remember there are certain colors that don't render correctly within `MaterialTheme3`, but I forget what the cases are

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowContainer.kt
Kokobe: Makes sense! Thanks for catching that

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowContainer.kt
Kokobe: I actually confirmed with @hwiatneeva about a month ago that this was the header text font we wanted! 
TitleMedium is too bold

Screen Shot 2023-01-24 at 10 49 43 AM

app/src/main/java/com/neeva/app/welcomeflow/login/WelcomeFlowNavModel.kt
Kokobe: `Activity.java` says no

Screen Shot 2023-01-24 at 10 52 18 AM

app/src/main/java/com/neeva/app/welcomeflow/login/LaunchLoginFlow.kt
Kokobe: done + `navigateToScreenAfterLoginWhenReady` renamed to `navigateToLoginResultScreenWhenReady`

app/src/main/java/com/neeva/app/welcomeflow/login/WelcomeFlowNavModel.kt
Kokobe: 1. It's a `WeakReference` not an actual activity reference, but I do see that I should rename it to `activityWeakReference`... This shouldn't have any memory-leak ramifications though right? The Garbage collector won't count these references as actual references to the Activity, so it can safely be garbage collected

Also I need this activityReference so I can call `finishAndRemoveTask()` which is an Activity-only method. 


2. Fair point. Okay I'll do that

app/src/main/java/com/neeva/app/welcomeflow/WelcomeFlowActivity.kt
Kokobe: https://github.com/neevaco/neeva-android/issues/1227

app/src/main/java/com/neeva/app/welcomeflow/login/WelcomeFlowNavModel.kt
dan-alcantara-neeva: That's not what that says.  Java automatically assumes nullable fields -- Kotlin requires you to say they're nullable or not.

app/src/main/java/com/neeva/app/welcomeflow/login/LoadingScreen.kt
dan-alcantara-neeva: Check the annotation definitions.  The Dark versions tell Compose to use a dark theme system environment, which sets various things to dark.

app/src/main/java/com/neeva/app/welcomeflow/login/LoadingScreen.kt
dan-alcantara-neeva: 1. Did you ask UX?  It looks like it was eyeballed to be centered on the screen.
2. You can do it as a follow up.  You can try swapping out the entire scrollable container with a loading Composable so that you can place it based on the whole container.

app/src/main/java/com/neeva/app/welcomeflow/login/WelcomeFlowNavModel.kt
dan-alcantara-neeva: WeakReference makes it memory-leak safe, but it's still bad encapsulation to store the Activity if you can avoid it.

app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -217,52 +218,52 @@ class SubscriptionManager(
...
+                return@launch
+            }
 
-                val obfuscatedUserID = billingClientController.obfuscatedUserIDFlow.value
-                // TODO(kobec): add proper error handling here and don't allow the purchase
-                //  if obfuscatedUserID == null.
-                //  https://github.com/neevaco/neeva-android/issues/1197
+            val offerDetails = retrieveEligibleOffers(
+                offerDetails = productDetails.subscriptionOfferDetails,
+                tag = tag.lowercase()
```
dan-alcantara-neeva, line 245: If you're always going to require that the tag is lowercase, move the `lowercase()` call to the `retrieveEligibleOffers` function.  Otherwise, you can get into a situation where you forget to do it at one callsite.

dan-alcantara-neeva:
> 

I was asking for a doc with an explanation for why you're doing what you're doing the way you did it (in English), plus a flow chart showing the different ways you can send the user around -- not a PR comment, a PR description, and a pointer to undocumented individual functions in a Model.

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
```diff
@@ -478,35 +520,109 @@ class FirstRunModel internal constructor(
...
+                    signup = launchLoginFlowParams.signup,
+                    provider = launchLoginFlowParams.provider,
+                    mktEmailOptOut = launchLoginFlowParams.mktEmailOptOut,
+                    loginHint = launchLoginFlowParams.emailProvided ?: ""
                 )
+            )
+        }
+
+        extractLoginUri(result, launchLoginFlowParams)?.let {
+            onGoogleSignInSuccess(it)
```
dan-alcantara-neeva, line 559: Why are you calling `onGoogleSignInSuccess` here if it can still fail and call `onGoogleSignInFailure` if the cookie doesn't get extracted?

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
```diff
@@ -478,35 +520,109 @@ class FirstRunModel internal constructor(
...
+        loginToken.updateCachedCookie(newLoginCookie)
+        // Since it is possible that NeevaActivity has not started yet, we need to make sure that
+        // the browser cookie jar has the new cached cookie value.
+        loginToken.updateBrowserCookieJarWithCachedCookie(
+            onFailure = {
+                // Tell the user something went wrong and that they couldn't be signed in.
+                popupModel.showSnackbar(appContext.getString(R.string.error_generic))
+            }
+        )
+        queueFetchNeevaInfo()
```
dan-alcantara-neeva, line 625: In a comment, explain why you're still fetching neeva info if the browser login cookie failed to be updated.

app/src/main/java/com/neeva/app/welcomeflow/login/WelcomeFlowNavModel.kt
```diff
@@ -34,26 +57,82 @@ class WelcomeFlowNavModel(private val navHost: NavHostController) {
...
+        val returnToActivityAndScreen = destinationToReturnWhenFinished.value != null
+
+        val browserIntent = activity.intent.apply {
+            setClass(activity, NeevaActivity::class.java)
+            flags = Intent.FLAG_ACTIVITY_NEW_TASK
+
+            when {
+                returnToActivityAndScreen -> {
+                    action = NeevaActivity.ACTION_SHOW_SCREEN_AFTER_LOGIN_SUCCESS
+                    destinationToReturnWhenFinished.value?.let {
```
dan-alcantara-neeva, line 123: Why don't you unset this value?

dan-alcantara-neeva (changes requested):

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
Kokobe: This is what happens when I don't immediately call onGoogleSignInSuccess(it)  -- which moves it to the loading screen  -- after getting the loginUri from GoogleSignIn

https://user-images.githubusercontent.com/20916043/214638465-e8678320-897b-4617-a6dc-b64e3f365020.mp4

Kokobe: In areas with super slow network connection, it can look like a successful native google sign in didn't actually do anything. I'll differentiate the process of extracting the cookie vs. getting the native google sign in uri  by renaming the lambdas

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
Kokobe: In this case, we don't care about the browser cookie jar failing to update because:
1. It will fail to update the cookie jar probably because the browser isn't alive yet.
2. it eventually will update when the browser comes to life. 

The comment I wrote though is:
`Fetch the user's Neeva information so we can see if they are allowed to buy premium.`

app/src/main/java/com/neeva/app/welcomeflow/login/WelcomeFlowNavModel.kt
Kokobe: ???

Kokobe: Offline convo: `destinationToReturnToWhenFinished` needed to be reset in finishWelcomeFlow() because if we reopened the WelcomeFlowActivity and it was still alive somehow, it would retain the old value of `destinationToReturnToWhenFinished`

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
dan-alcantara-neeva: Would prefer you put that context in the code

dan-alcantara-neeva (approved):
LGTM % last comment about comment

PR #1228, opened 2023-01-24
SingleTabActivity: Fix illegal state exception when attachWebLayerBrowserFragment

Kokobe:
We were getting the following crash stack on release (0.96.6): https://play.google.com/console/u/0/developers/6544928132232754928/app/4975546713093074924/vitals/crashes/1ef0c51d9041b308ae26dfc7f1885440/details?installedFrom=PLAY_STORE&days=28

```
Exception java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
  at androidx.fragment.app.FragmentManager.checkStateLoss (FragmentManager.java:3)
  at androidx.fragment.app.FragmentManager.execSingleAction (FragmentManager.java)
  at androidx.fragment.app.BackStackRecord.commitNow (BackStackRecord.java:2)
  at com.neeva.app.singletabbrowser.SingleTabActivityViewModel.attachWebLayerBrowserFragment (SingleTabActivityViewModel.java:8)
  at com.neeva.app.singletabbrowser.SingleTabActivityViewModel$initializeWebLayer$1.invokeSuspend (SingleTabActivityViewModel.java)
  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (BaseContinuationImpl.java)
  at kotlinx.coroutines.DispatchedTask.run (DispatchedTask.java)
  at android.os.Handler.handleCallback (Handler.java:938)
  at android.os.Handler.dispatchMessage (Handler.java:99)
  at android.os.Looper.loop (Looper.java:268)
  at android.app.ActivityThread.main (ActivityThread.java:7904)
  at java.lang.reflect.Method.invoke (Method.java)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:627)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:997)
```

Also:
* Adds check for `isBrowserAlive` to avoid using the browser outside of onCreate and onDestroy (another crash we found). 

## What is happening?

When `transaction.commitNow()` is called, it could be called after `Fragment.onSaveInstanceState`. According to the `FragmentManager.beginTransaction()`'s kdocs:

Screen Shot 2023-01-25 at 12 43 52 AM

So in situations where we cannot `commitNow` and it throws an IllegalStateException, it means that the browser is not ready to be used. 

We put a try catch around the commitNow() and returned a null fragment to signify an error. 

* https://play.google.com/console/u/0/developers/6544928132232754928/app/4975546713093074924/vitals/crashes/1ef0c51d9041b308ae26dfc7f1885440/details?installedFrom=PLAY_STORE&days=28

Screen Shot 2023-01-25 at 12 49 30 AM

### How was this tested?
- [ ] I added automated tests

### Manual test cases
Turned on the debug flag to have the OS garbage collect any app I currently cannot see. Saw that SingleTabActivity did not crash. 

https://user-images.githubusercontent.com/20916043/214357032-d1cd2ee1-8bc8-42f7-ba31-d3554520d291.mp4

Happy Path: Opened SingleTabActivity. Nothing crashed.

https://user-images.githubusercontent.com/20916043/214356627-a19b602d-cfe2-412a-9c8f-ca81aa92b72f.mp4


app/src/main/java/com/neeva/app/singletabbrowser/SingleTabActivityViewModel.kt
```diff
@@ -97,7 +99,12 @@ class SingleTabActivityViewModel @Inject constructor(
                 }
         }
 
-        transaction.commitNow()
+        try {
+            transaction.commitNow()
```
dan-alcantara-neeva, line 103: Seems like the only real difference between the two is that `commitAllowingStateLoss` doesn't throw an exception.  Might as well just use that directly instead of wrapping it in a try/catch.

dan-alcantara-neeva: Debugged in a pair programming session: We need `commitNow` because the callsite needs to know if the `BrowserFragment` is available immediately.  Instead, we'll add this try/catch and then confirm that the Activity is alive before trying to use the Browser.

dan-alcantara-neeva (approved):

PR #1239, opened 2023-01-31
Another Speculative fix for the showBrowser AppNavModel Crash

Kokobe:
## What is the problem?
See stack trace in #939 

The method `showBrowser` makes a call to `appNavModel?.showBrowser` which uses a `NavHostController` that is a null pointer. A NPE is triggered from this. 

## How can this happen?

My theory: it seems that when the app gets killed by the OS (for taking up too much memory and not being in the foreground anymore), the `NavHostController` pointer will point to `null` because the UI has been cleaned up out of main memory. 

When the app `RESUMES` and gets recreated, `registerBrowserCallbacks()` -- a method that can happen before UI gets constructed -- eventually triggers `NeevaActivity's` `showBrowser()` . This method of course is using a `navHostController` that points to null. 

Although we have a check to make `showBrowser()` wait for the `firstComposeCompleted`, it MIGHT be calling `complete(true)` in the wrong place and with the wrong parameters. See below:

```
// ActivityUI and Neeva theme above
LaunchedEffect(true) {
       firstComposeCompleted.complete(true)
}
``` 
It might not work because:
* It doesn't launch if the `navHostController` changes
* It isn't a disposable effect that is tied directly to the UI (has no way of saying the activity actually doesn't work rn
* It isn't being called right after `AnimatedNavHost()` (more on this in the next section)

## Why would this PR work?
* According to https://stackoverflow.com/questions/64198980/nullpointerexception-when-trying-to-get-navcontroller the NavController can only be used once the UI is ready. 

* My own experience with NPE `NavHostController`s when implementing the `WelcomeFlow`, it seems that the `NavHostController` only becomes a non-null pointer once `AnimatedNavHost` in `AppNav.kt` has finished executing. 

* Also by using a `DisposableEffect`, we can set `isNavHostReady` directly based on whether the UI is ready or killed (I checked using logs). 

* Issue we've been trying to fix forever:  #939 

### How was this tested?
- [ ] I added automated tests

### Manual test cases
Added logs for when the `isNavHostReady` flow changes + when `appNavModelImpl?.showBrowser()` is called. 
Confirmed that  when `isNavHostReady` is true is always followed by `appNavModelImpl?.showBrowser()` getting called. 

## Issues addressed
*  #939


app/src/main/java/com/neeva/app/NeevaAppModules.kt
```diff
@@ -205,6 +206,12 @@ object NeevaAppModule {
         )
     }
 
+    @Provides
+    @Singleton
+    fun provideNavControllerReadyModel(): NavControllerReadyModel {
```
Kokobe, line 211: Don't need this, just pass it down

app/src/main/java/com/neeva/app/NeevaActivity.kt
```diff
@@ -330,12 +335,22 @@ class NeevaActivity : AppCompatActivity(), ActivityCallbacks {
...
-            // Speculative fix for https://github.com/neevaco/neeva-android/issues/939
-            // Wait until the first Compose has completed, which should mean that the NavGraph has
-            // been set up with all of the destinations.
-            firstComposeCompleted.await()
-            appNavModel?.showBrowser(forceUserToStayInCardGrid = forceUserToStayInCardGrid)
+        lifecycleScope.launch(dispatchers.io) {
+            // Another speculative fix https://github.com/neevaco/neeva-android/issues/939
+            // It seems that the NavHostController will no longer be a null pointer only after the
+            // AnimatedNavHost() function has finished executing.
+            navControllerReadyModel.isNavHostReady
```
Kokobe, line 342: use completable deferred await()

app/src/main/java/com/neeva/app/ActivityUI.kt
```diff
@@ -76,7 +77,8 @@ val LocalWelcomeFlowNavModel = compositionLocalOf<WelcomeFlowNavModel> { error("
 @Composable
 fun ActivityUI(
     toolbarConfiguration: StateFlow<ToolbarConfiguration>,
-    webLayerModel: WebLayerModel
+    webLayerModel: WebLayerModel,
+    isNavControllerPointerNonNull: CompletableDeferred<Boolean>
```
dan-alcantara-neeva, line 81: Eh... `isNavControllerAvailable`?

app/src/main/java/com/neeva/app/appnav/AppNav.kt
```diff
@@ -153,4 +155,13 @@ fun AppNav(
 
         signInFlowNavGraph()
     }
+
+    // At this point, we know that the nav host has definitely been created and is not a
+    // null pointer.
+    DisposableEffect(appNavModel.navController) {
+        isNavControllerPointerNonNull.complete(true)
+        onDispose {
+            // No cleanup necessary. When the activity restarts, the above state should re-init too.
```
dan-alcantara-neeva, line 164: Cleanup would be necessary if the AppNav is recomposed without the app dying.

Kokobe: good point, going to use a MutableStateFlow instead then

app/src/main/java/com/neeva/app/appnav/AppNav.kt
```diff
@@ -41,6 +42,7 @@ fun AppNav(
     toolbarConfiguration: StateFlow<ToolbarConfiguration>,
     webLayerModel: WebLayerModel,
     appNavModel: AppNavModel,
+    isNavControllerPointerUnavailable: MutableStateFlow<Boolean>,
```
dan-alcantara-neeva, line 45: Avoid building in an implicit `not` into a variable name like this -- especially with the negation is "un".  Call it `isNavControllerPointerAvailable`

Kokobe: Agh good catch, turns out I named the variable wrong and I was using it as `isNavControllerPointerAvailable`

dan-alcantara-neeva (approved):

PR #1244, opened 2023-02-01
Add hooks for detecting redirects that the client needs to handle

dan-alcantara-neeva:
https://neevaco.slack.com/archives/C02RYG8LY8H/p1675193084393309

When doing the native google sign in, we need to detect when the backend redirects us to a `neeva://` URI.  This can indicate that an error code was sent back that we need to handle.


app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
```diff
@@ -571,39 +573,72 @@ class FirstRunModel internal constructor(
         }
     }
 
+    private fun connectAndFollowRedirects(uri: String, numFollowsAllowed: Int = 5): String? {
+        val CONNECTON_TIMEOUT_MILLISECONDS = 60000
+        var urlConnection: HttpURLConnection? = null
+        try {
+            val url = URL(uri)
+            android.util.Log.e("dfalcantara", "Opening $url")
```
clischeske, line 581: Do you want to leave these in?

clischeske (approved):
As far as I can tell it looks right

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
dan-alcantara-neeva: Yeah, it's just debugging logs.  Not expecting to submit this code because Kobe needs to hook into it

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
```diff
@@ -571,39 +573,72 @@ class FirstRunModel internal constructor(
...
+
+            // Connect and read the content of the page, but don't use it.
+            urlConnection.content
+
+            if (urlConnection.responseCode == HTTP_SEE_OTHER && numFollowsAllowed > 0) {
+                val redirectUrl = urlConnection.getHeaderField("Location")
+                val redirectUri = Uri.parse(redirectUrl)
+                val loginCallbackIntentParams = LoginCallbackIntentParams.fromUri(redirectUri)
+                return if (loginCallbackIntentParams != null) {
+                    android.util.Log.e("dfalcantara", "Detected neeva URI redirect, which the Android client is supposed to handle.  Error code on the URI: ${loginCallbackIntentParams?.retryCode}")
```
Kokobe, line 602: Documenting: httpURLConnection cannot handle the neeva uri

Kokobe: TODO: will need to handle error

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
```diff
@@ -571,39 +573,72 @@ class FirstRunModel internal constructor(
...
-                if (loginCookie != null) {
-                    loginCookieFound = true
-                    onSuccess(loginCookie)
-                }
-            } catch (e: Exception) {
-                Timber.e(e, "Error when trying to extract login cookie from Google uri.")
-            } finally {
-                urlConnection?.disconnect()
+            val finalUrl = connectAndFollowRedirects(uri.toString())
+            android.util.Log.e("dfalcantara", "Final URL: $finalUrl")
```
Kokobe, line 629: If a neeva url was returned, handle error + show error popup

PR #1256, opened 2023-02-05
Queue Neeva User Info Fetch after Billing updates

Kokobe:
Queues the fetch for neeva user info after a successful acknowledgement of a purchase. 

* https://www.figma.com/file/dylvDPB69JG9U1rzZ3HQ52/Android-Premium?node-id=0%3A1&t=oGyE4dBu3wuZBgaH-0

### How was this tested?
- [ ] I tested this manually on `debug`
- [ ] I added automated tests

### Manual test cases
Got a subscription acknowledged -> should see that Settings is immediately updated after a purchase. 

Cancelled a subscription. After the subscription period ends -> should see that the user is now on the Basic plan

Got a subscription acknowledged -> should see that Settings is immediately updated after a purchase. 

Cancelled a subscription. After the subscription period ends -> should see that the user is now on the Basic plan


dan-alcantara-neeva:
@Kokobe You missing the screenshots and videos here?

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
```diff
@@ -400,7 +408,21 @@ class FirstRunModel internal constructor(
         )
     }
 
-    private fun queueFetchNeevaInfo() {
+    private fun startPurchaseAcknowledgedConsumer() {
+        coroutineScope.launch(dispatchers.io) {
+            val purchaseAcknowledgedRecently = billingClientController.acknowledgePurchaseManager
+                .purchaseAcknowledgedRecently
+            // Once a purchase has been acknowledged, update the neeva user info:
+            purchaseAcknowledgedRecently
```
dan-alcantara-neeva, line 416: The way this is set up is a little odd.  You're basically using another class to store a variable that this class unsets and filters out so that it doesn't fire again.  That means that you just want to be alerted when something happens, which means you're basically setting up a callback.  In general, you should avoid breaking encapsulation by having other classes change other class' state.  That's why we use the `private MutableStateFlow` + `public StateFlow` pattern elsewhere.

Aside from that, there's a couple of ways you might want to set this up -- if this is the only consumer of that boolean.

1. **Use a timestamp instead of a boolean**
Instead of setting true/false, you can make it so that the `AcknowledgePurchaseManager` has a StateFlow of timestamps when it last acknowledged a purchase, starting with `null` if no purchase has been acknowledged recently.  `FirstRunModel` would still collect the flow, filtering out any non-null values.  Because the `StateFlow` prevents emission of the same values, you can be sure that your `collect` fires only once per acknowledgement.  You then don't have to unset the timestamp at all because you're working with `Singleton`s.

2. ** Set up a callback mechanism (less preferable because it makes the dependency graph messy) **
You can have the `AcknowledgePurchaseManager` manage its boolean state on its own and also maintain a callback.
* When it detects that a callback has been added, you would check the boolean and fire the callback immediately then unset the boolean.  The boolean in this case would effectively be caching an event for when it needs to be consumed.
* When it detects that a callback exists when a purchase has been acknowledged, the boolean doesn't get set and the callback is alerted directly.

dan-alcantara-neeva (changes requested):

app/src/main/java/com/neeva/app/firstrun/FirstRunModel.kt
Kokobe: I agree with the timestamp method, thank you!

Kokobe:
Going to convert to a draft because Premium cancellation changes are not working at the moment

app/src/main/java/com/neeva/app/billing/billingclient/AcknowledgePurchaseManager.kt
```diff
@@ -117,6 +121,8 @@ class AcknowledgePurchaseManager(
                     fetchContinuation.safeResumeContinuation(
                         BillingClientJobResult(success = success)
                     )
+
+                    _purchaseLastAcknowledgedAt.value = System.currentTimeMillis().toString()
```
dan-alcantara-neeva, line 125: Eh, just use a long instead of converting to a string?

dan-alcantara-neeva (approved):
lgtm % comment

PR #1264, opened 2023-02-16
Fix FirstRun ZeroQuery showing up all the time

### LINKED ISSUE #1236: Neeva ad-blocker promo appears again after going back to Fast Tap when closing a tab
**Describe the bug**
Although we should only show the promo once, the user can be shown the promo again.

**To Reproduce**

https://user-images.githubusercontent.com/96092902/215376909-acd3ed0b-86b8-41bc-95a3-36273d343abf.mp4

1. Click on the URL bar
2. Enter a query, but don't but enter
3. Tap on one of the presented results
4. Hit "Back" to get back to Fast Tap and the search results
5. Clear the query by hitting the x button in the query box
6. You'll end up back on Zero Query but it is the version you get when you go through first run.

**Expected behavior**
You should be presented with no promo, and should see the correct version of Zero Query.

**Affected version**
Version 103

Kokobe:
fixes #1236 

* From Hwi: The user should never see the FirstRun ZeroQuery again after using it for the first time. 

We needed a way to see if the user has searched before on the Neeva browser to do what Hwi wants above^^. However, navigations around Neeva `settings` on the website should not count as a "search navigation". 

So I introduced a term called "Search Navigation" which is:  
- any website that is a search within the neeva domain
- any website that is outside of the neeva domain (for example, the user just typed "google.com" into the urlbar and hit enter)

What doesn't count as a "Search Navigation" is:
- anything that is not a search within the neeva domain
  - example: navigating around the neeva website and looking at settings

If the user has more than 0 Search Navigations, then that means they must have used FirstRun ZeroQuery.

### How was this tested?
- [ ] I added automated tests

### Manual test cases

Only "Search Navigations" (see `HistoryDao.kt` for what defines this) will set `FirstSearchDone` to true:
* Wiped app memory
* Started FirstRun
* Saw FirstRun ZeroQuery
* Navigated to Account Settings 
* Saw that ZeroQuery is still FirstRun (because a search navigation has not been made yet)
* Uses FirstRun ZeroQuery
* Sees that ZeroQuery is no longer showing the FirstRun version.

https://user-images.githubusercontent.com/20916043/219263289-0d137e1a-2dad-42b1-8e84-71db0881229f.mov

Using the FirstRun ZeroQuery from the start also hides it for the rest of the user's usage of the app:

https://user-images.githubusercontent.com/20916043/219263087-7aaa8e26-647e-4ac2-94ce-1eda208f86d6.mov

## Issues addressed
* Fixes #1236


app/src/main/java/com/neeva/app/ui/BrowserScaffold.kt
```diff
@@ -205,7 +206,8 @@ private fun BoxScope.BrowserOverlay(
     val toolbarConfiguration by toolbarConfigurationFlow.collectAsState()
     val urlBarModelState: URLBarModelState by urlBarModel.stateFlow.collectAsState()
     val isEditing = urlBarModelState.isEditing
-    val focusUrlBar = urlBarModelState.focusUrlBar
+    val isFirstSearchDone = SharedPrefFolder.FirstRun.FirstSearchDone
```
yasminyin96, line 209: Maybe `val isFirstSearchDone by SharedPrefFolder.FirstRun.FirstSearchDone.getFlow(sharedPreferencesModel).collectAsState()`?

yasminyin96 (approved):

PR #1300, opened 2023-03-02
Adds sign in options whenever Native Google Sign In is used

Kokobe:
Because the `GoogleSignInClient` caches the last signed in account (to make the sign in process faster), it only allows you to sign in with that one account after a first sign in. The only way to manually make the client give you the option of choosing an email account is by removing that account from your device / Play Store.

This is an issue for people with multiple google accounts on Play Store because in order to sign in with a different account, they need to wipe device state and do extra signing in that they might not know they had to do in the first place. The Google dev docs recommends to call this when the user logs out: https://developers.google.com/identity/sign-in/android/disconnect

I put it here instead of where the signOut logic is because there could be an edge case that we sign the user out but `GoogleSignInClient.signOut()` never gets called. If something crashes in the sign out process - i.e.  maybe chromium crashes the app when updating the cookie jar or there's some other bug, `GoogleSignInClient.signOut()` never gets called and then the user is stuck on signing in with that one email account. 
```
// Sign out code:
neevaUser.clearUserInfo()
googleSignInClient.signOut()
```

I put `googleSignInClient.signOut()` before every login because I'm coding defensively here. By making sure that the Native Google Sign In flow is always going to be consistent when its launched, it will always give users the option to choose an account when signing in. 

### How was this tested?
- [ ] I added automated tests

## Tested Sign in on multiple accounts on the Prod server:

Checks that sign in correctly updates the browser cookie and the Profile information when signing in with multiple accounts:

https://user-images.githubusercontent.com/20916043/222530790-c5b07b56-da08-455d-b7be-5b8078db2920.mp4


Kokobe:
@dan-alcantara-neeva make sure to update the Gist app's native google sign in wrapper code

app/src/main/java/com/neeva/app/firstrun/nativelogin/NativeGoogleSignInWrapper.kt
```diff
@@ -88,6 +88,7 @@ open class NativeGoogleSignInWrapper constructor(
             .requestEmail()
             .build()
         googleSignInClient = GoogleSignIn.getClient(activityContext, signInOptions)
+        googleSignInClient.signOut()
```
dan-alcantara-neeva, line 91: Why is signing the user out before they sign in the correct place to do this?

dan-alcantara-neeva: Also, why are you ignoring the Task it provides to see if it succeeded or fails?  This is an async operation.

dan-alcantara-neeva:
Also, use your PR description to describe the other things you tried and why this is the right place to put the one liner.  All it says it what the effect of the PR is but not any of the reasoning behind it.

app/src/main/java/com/neeva/app/firstrun/nativelogin/NativeGoogleSignInWrapper.kt
Kokobe: Because the `GoogleSignInClient` caches the last signed in account (to make the sign in process faster), it only allows you to sign in with that one account after a first sign in. 

This is an issue for people with multiple google accounts on Play Store because in order to sign in with a different account, they need to remove the others from their device/PlayStore. Google dev docs recommends to call this when the user logs out: https://developers.google.com/identity/sign-in/android/disconnect

I put it here instead of where the signOut logic is because there could be an edge case that we sign the user out but `GoogleSignInClient.signOut()` never gets called. If something crashes in the sign out process - i.e.  maybe chromium crashes the app when updating the cookie jar or there's some other bug, `GoogleSignInClient.signOut()` never gets called and then the user is stuck on signing in with that one email account. 
```
// Sign out code:
neevaUser.clearUserInfo()
googleSignInClient.signOut()
```

I'm coding defensively here by making sure that the Native Google Sign In flow is always going to be consistent and giving users account options when signing in. 

I'll update the PR description + add a comment in the code with this reasoning.

dan-alcantara-neeva: You need to respond to the second comment.

Kokobe: Sorry missed it because I didn't refresh my page! 

Yes you are correct, I should have used the `addOnCompleteListener` from the Google API

dan-alcantara-neeva (approved):

PR #1301, opened 2023-03-03
Fix navigation logic and billing flow launch logic

Kokobe:
Although the WelcomeFlowActivity has the "removeFromRecents" flag set to `true` in `AndroidManifest`, this does not mean that the activity will die when it is removed from Recents. This is an assumption I made when writing the `WelcomeFlowNavModel` and the launchBillingFlow logic (see: the end of `private fun buy()` in `SubscriptionManager.kt`)

There were 2 problems that were caused by that assumption and fixed by this PR.

## Problem 1) The Billing Flow was able to launch while the `WelcomeFlowActivity` **was not visible**

As a result, when you opened the `WelcomeFlowActivity` from Settings and selected a `Premium` subscription, it queued a job -- that will run when the user signs in -- and launch the billing flow. Now if you were to login from a tab in the Browser in `NeevaActivity`, it will launch the `BillingFlow` and put the `WelcomeFlowActivity` on top. 

### Solution
Just as the `private fun buy()` function in `SubscriptionManager` checks if the `WelcomeFlowActivity` is alive before launching the BillingFlow, now I am tracking and checking another activity lifecycle state: `isInForeground`. This is a member variable of `WelcomeFlowActivity` and is set when:
* onResume -> `isInForeground` set to `true`
* onStop -> `isInForeground` set to `false`

Now, the `BillingFlow` won't be unexpectedly launched if the `WelcomeFlowActivity` is not on visible. 

## Problem 2) Because the `WelcomeFlowActivity` doesn't die, the old back stack is still saved
When launching the `WelcomeFlowActivity` via Intent, the navigation back-stack for that activity was not being popped/reset when we set the `startDestination`. 

* `popUpTo()` does not actually do anything if the destination does not exist in the back stack.
* Instead, when navigating to a `startDestination`, we will manually pop everything from the back stack and then navigate to the startDestination. (see: `resetBackStackAndShow()` in `WelcomeFlowNavModel.kt`)
* I also made sure that each `composable` Destination in the navigation host has a unique name to ensure that `popUpTo` can work properly

## Small login change:
* `firstRunModel.setLoginAttemptOccurred(true)` was being used for `SingleTabActivity` logins to induce a `Loading` state after a login attempt. These were used in `processNewIntent()` and `processCreateIntent()` in `WelcomeFlowNavModel`
* Native Google Login was wrongly calling `firstRunModel.setLoginAttemptOccurred(true)` when its `Loading` state was already being handled by the `NativeGoogleSignInWrapper`

* https://www.figma.com/file/dylvDPB69JG9U1rzZ3HQ52/Android-Premium?node-id=0%3A1&t=d7w7oX3x6rjJVA5o-0

### How was this tested?
- [ ] I added automated tests

#### Billing Flow not Launched on Sign In from Tab + navigation back stack is reset: 
* Selected a Premium Plan in WelcomeFlow from Settings (queues the Billing Flow job)
* Exits WelcomeFlowActivity and logs in from a tab in NeevaActivity
* See that BillingFlow is not launched!
* Go to user Profile in Settings and open "Learn more about Premium"
* See that the navigation back stack has been reset for WelcomeFlowActivity (clicking back does not show the older screens)

https://user-images.githubusercontent.com/20916043/222816831-4db7c16a-915f-4ec1-8284-96cffc11a139.mp4

#### Above test but on "Don't keep Activities" option on

https://user-images.githubusercontent.com/20916043/222945570-2795c124-eeb6-465c-b0d4-4481bd2a2e21.mp4

#### FirstRun, Billing Enabled

https://user-images.githubusercontent.com/20916043/222945700-43b36c94-8ea4-4fd3-9391-aa31a1055827.mp4

#### FirstRun, Billing Enabled "Don't keep Activities"
- It opens settings right after FirstRun because the last screen I was on was Settings when I selected the dev flag"Simulate First Run"

https://user-images.githubusercontent.com/20916043/222945739-145a89f8-4852-4569-8072-087d73e3be00.mp4

#### FirstRun, Billing Enabled (Custom Tab) 

https://user-images.githubusercontent.com/20916043/222946951-849eba52-ba97-4c09-bc6b-5ca97d01d32a.mp4

#### FirstRun, Billing Enabled (Custom Tab) "Don't keep Activities"

https://user-images.githubusercontent.com/20916043/222946909-a9467358-a519-4df5-98a2-b8ed77fe4cfb.mp4

#### FirstRun, Billing Disabled -> (WelcomeFlowActivity is set properly)

https://user-images.githubusercontent.com/20916043/222925615-635e021e-a7cb-4bb2-9f51-7f82648355e4.mp4


app/src/main/java/com/neeva/app/billing/SubscriptionManager.kt
```diff
@@ -288,6 +289,9 @@ class SubscriptionManager(
                 // If the WelcomeFlowActivity is dead, there is no reason to call
                 // onBillingFlowFinished().
```
dan-alcantara-neeva, line 290: Update the comment to say that you shouldn't do this if the app isn't in the foreground, either.

app/src/main/java/com/neeva/app/welcomeflow/login/WelcomeFlowNavModel.kt
```diff
@@ -75,21 +78,55 @@ class WelcomeFlowNavModel(
 
     internal fun show(destinationName: String, setOptions: NavOptionsBuilder.() -> Unit = {}) {
         coroutineScope.launch(dispatchers.main) {
+            // Wait for the navHostController to be initialized
```
dan-alcantara-neeva, line 81: Pull this out into a function

app/src/main/java/com/neeva/app/welcomeflow/login/WelcomeFlowNavModel.kt
```diff
@@ -75,21 +78,55 @@ class WelcomeFlowNavModel(
...
+    ) {
+        coroutineScope.launch(dispatchers.main) {
+            // Wait for the navHostController to be initialized
+            isNavHostControllerAvailable
+                .filter { it }
+                .first()
+
+            navHost?.let { navHostController ->
+                // Reset the back stack by clearing everything
+                navHostController.backQueue
```
dan-alcantara-neeva, line 114: I think you're getting lucky here:
`popBackStack` removes everything on top of the thing that you popped, which will probably mess with the `backQueue` you're iterating through.  I think you're only getting lucky and dodging a ConcurrentModificationException because `mapNotNull` creates a new copy of the array.  Can you make that explicit?

Alternatively, is there a better way to do this?

Kokobe: Went with:
```
// Reset the back stack by popping everything
while (navHostController.backQueue.any { it.destination.route != null }) {
   navHostController.popBackStack()
}
```

Kokobe: This was bad because it's blindly popping without taking any specific Destinations in the backQueue into account. 

Going with a different method

dan-alcantara-neeva (approved):

PR #1307, opened 2023-03-11
NativeGoogleSignIn: Properly use result cancelled code exit

Kokobe:
Before because of #1304, we were wrongly opening up SingleTabActivity when the user cancels out of the Native Google Sign In Activity. This is the behavior we would get:

https://user-images.githubusercontent.com/20916043/224459009-a0fd0372-b087-4c85-a8c1-2bec21b67acc.mp4

### How was this tested?
- [ ] I added automated tests

### Manual test cases

https://user-images.githubusercontent.com/20916043/224458965-0cfed8ea-b92a-402b-9bba-22a6f60b3f55.mp4

#### When Native Google Sign In fails (on an unsigned release build):

https://user-images.githubusercontent.com/20916043/224827024-f85022c9-c29c-4478-ba10-af3093b9ea51.mp4


Kokobe:
Because the Google Native Login API will return a Login API Error + return a `result` with the `resultCode` = `RESULT_CANCELLED` -- even if the user simply dismisses the prompt -- we are given no signal to determine if there was an actual API error or the user just dismissed the prompt. 

It seems like most apps (like Evergreen) don't open up a custom tab in case the Native Google Sign In API fails, which is why they avoid the weird scenario in the first video of the PR description. Since we're deciding to give the user another option to sign in (in the case the the Native Google Sign In API fails and sends a Login API error), this UX is the tradeoff we are consciously making.

PR #1313, opened 2023-03-15
Use Feature Flag from UserInfoQuery for deciding isPremiumAvailable

Kokobe:
There is a feature flag on the back-end that we can read from to disable Billing remotely (see: https://github.com/neevaco/neeva/pull/97496)

This PR utilizes this flag and does not allow any purchases if the billing feature flag is disabled. Since the feature flag is only in the `UserInfoQuery`, we won't be able know billing is unavailable until the user signs in. 

## Expected Behavior
Before the user signs in, there is no way of us getting the enable billing feature flag because it is tied to the `UserInfoQuery`. The default behavior is to show them the PLANS screen and direct them to login after selecting a plan. 

Once the user is logged in, the enable billing feature flag can either exist or not exist.
* If the flag exists, we will allow Billing to happen and say it is available (see SubscriptionManager)
* If the flag does not exist (because the server is screwed up) or its value is false, then we will say that Premium is unavailable (the billing flow will never get launched). 

## Potential Pitfalls
If we were to get the following scenario:
* the user has logged in from a previous session
* the UserInfo is cached (along with the feature flag that says it billing is enabled)
* the neeva server is not returning the UserInfoQuery but the flag has been disabled

The user would be able to access the PlansScreen and launch the billing flow. Now if they were to make a purchase, it would only get acknowledged if `RegisterGooglePlaySubscription` (another back-end call) succeeds even if the feature flag is turned off. I'll get in touch with @clischeske to ensure this doesn't happen, but the code logic in this snippet creates the correct behavior.

### The above scenario can also only happen if app start-up has not reset the `UserInfo`
If the fetchUserInfo query (which happens on start-up) returns an error or times out, it is set to clear all of the user info --> which prevents any of the above scenario from happening because they would need to sign in to make a purchase.

### How was this tested?
- [x] I added automated tests

#### Tested with the back-end feature flag disabled:

https://user-images.githubusercontent.com/20916043/225480339-db00633a-09d1-4bac-acef-08d2426d6b3c.mp4

#### Tested with the back-end feature flag enabled:

https://user-images.githubusercontent.com/20916043/225486490-023f0fab-cdac-4e24-beae-843b0a3df658.mp4


app/src/main/java/com/neeva/app/userdata/NeevaUser.kt
```diff
@@ -246,7 +252,8 @@ class NeevaUserImpl(
                     SubscriptionSource.values()
                         .firstOrNull { it == subscription?.source }
                         ?: SubscriptionSource.UNKNOWN__
-                    ).name
+                    ).name,
+                isBillingEnabledInBackEnd = isBillingEnabled == true
```
dan-alcantara-neeva, line 256: "backend" is one word, so don't camelcase the End

dan-alcantara-neeva: Did you check with Carl about whether this flag should be enabled or disabled by default?

app/src/main/java/com/neeva/app/userdata/NeevaUser.kt
```diff
@@ -227,6 +229,10 @@ class NeevaUserImpl(
         private val FETCH_COOLDOWN = TimeUnit.MINUTES.toMillis(1)
 
         fun UserInfoQuery.User.toUserInfo(): UserInfo {
+            val isBillingEnabled = featureFlags
```
dan-alcantara-neeva, line 232: Instead of filtering out and throwing out the flags, can you save them to the `UserInfo` so that we can display them in the developer settings, or query them for other flags later?

app/src/main/java/com/neeva/app/userdata/NeevaUser.kt
Kokobe: It was disabled by default (https://github.com/neevaco/neeva/pull/97496), but now it is enabled by default (https://github.com/neevaco/neeva/pull/97614)

dan-alcantara-neeva: That's only if the flag is even fetchable.  What about if you can't talk to the server?

app/src/main/java/com/neeva/app/userdata/NeevaUser.kt
dan-alcantara-neeva: https://github.com/neevaco/android-llm/pull/230

app/src/main/java/com/neeva/app/userdata/NeevaUser.kt
Kokobe: Before the user signs in, there is no way of us getting the enable billing feature flag because it is tied to the `UserInfoQuery`. The default behavior is to show them the PLANS screen and direct them to login after selecting a plan. 

Once the user is logged in, the enable billing feature flag can either exist or not exist.
* If the flag exists, we will allow Billing to happen and say it is available (see SubscriptionManager)
* If the flag does not exist (because the server is screwed up) or its value is false, then we will say that Premium is unavailable (the billing flow will never get launched). 

Now if we were to get the following scenario:
* the user has logged in from a previous session
* the UserInfo is cached (along with the feature flag that says it billing is enabled)
* the neeva server is not returning the UserInfoQuery but the flag has been disabled

The user would be able to access the PlansScreen and launch the billing flow. Now if they were to make a purchase, it would only get acknowledged if `RegisterGooglePlaySubscription` (another back-end call) succeeds even if the feature flag is turned off. I'll get in touch with carl to ensure this doesn't happen, but the code logic in this snippet creates the correct behavior.

Kokobe: Also on app start-up, if the fetchUserInfo query returns an error or times out, it is set to clear all of the user info

dan-alcantara-neeva (approved):

